PR #86, opened 2015-02-17
Fixed RecurringApplicationCharges resources current method (Issue: 85)

mentholi:
This pull request fixes issue https://github.com/Shopify/shopify_python_api/issues/85 where RecurringApplicationChrage's current() method didn't work as intended. Intend of the method was to easily find resource with status = "active" but as the api end point doesn't support filtering by that field any other resource could be returned.

This is now fixed by doing filtering in memory and returning resource with correct status or None if not found. 
```python
def _get_first_by_status(resources, status):
    for resource in resources:
        if resource.status == status:
            return resource
    return None


class RecurringApplicationCharge(ShopifyResource):

    @classmethod
    def current(cls):
        """
        Returns first RecurringApplicationCharge object with status=active.
        If not found, None will be returned.
        """
        return _get_first_by_status(cls.find(), "active")
```

Behaviour is similar to Ruby api. See: https://github.com/Shopify/shopify_api/blob/08a015701d828cc85e38eb5a232bbcfd54518983/lib/shopify_api/resources/recurring_application_charge.rb#L7

PR #89, opened 2015-02-26
Update MD5 Signature validation to HMAC

### LINKED ISSUE #84: Update MD5 Signature validation to HMAC
Docs say that MD5 signature validation 'To be removed after June 1st, 2015'

It appears ```Session``` is still doing it this way vs HMAC

gavinballard:
This PR replaces the soon-to-be-deprecated MD5 method of OAuth validation with the HMAC-SHA256 method described [in the Shopify OAuth documentation](http://docs.shopify.com/api/authentication/oauth#verification).

It includes the addition of a few tests to verify that the HMAC calculation is being performed correctly (based on the examples in the Shopify documentation) and replaces tests verifying the calculation of the `signature` parameter with their `hmac` equivalents.

Has been tested in a live production environment without any problems. This PR closes #84.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,7 @@
 
 * Added Checkout resource
 * Updated to pyactiveresource v2.1.1 which includes a test-related bugfix
+* Changed OAuth validation from MD5 to HMAC-SHA256
 
 == Version 2.1.0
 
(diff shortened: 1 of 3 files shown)


gavinballard:
The build failures here are the same as those mentioned in PR #86 (issue with `pyactiveresource` and Python v2.7.9).

gavinballard:
@kevinhughes27 Ping!

PR #91, opened 2015-02-27
Fix for prefix issue in base_test.py

gavinballard:
After [this change](https://github.com/Shopify/pyactiveresource/pull/12) to pyactiveresource, `test_delete_should_send_custom_headers_with_request` was failing due to inconsistent expected and actual URLs. This failure was causing other tests to fail as `shopify.ShopifyResource.set_headers` weren't being reset after the failure.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,6 +1,7 @@
 == Version 2.1.1 (unreleased)
 
 * Added Checkout resource
+* Updated to pyactiveresource v2.1.1 which includes a test-related bugfix
 
 == Version 2.1.0
 
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ setup(name=NAME,
       scripts=['scripts/shopify_api.py'],
       license='MIT License',
       install_requires=[
-          'pyactiveresource>=2.1.0',
+          'pyactiveresource>=2.1.1',
           'PyYAML',
           'six',
       ],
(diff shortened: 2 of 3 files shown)


mentholi:
Hello @gavinballard! 

I can confirm that this fixes errors on tests with Python 2.7.9. I confirmed that by running tests against your branch https://github.com/gavinballard/shopify_python_api/tree/prefix_fix using newest master of https://github.com/Shopify/pyactiveresource which now contains your another fix. 

Travis build is failing for this branch because it still uses pyactiveresource without your fix https://github.com/Shopify/pyactiveresource/pull/13.

I think it would be good idea to update version number of pyactiveresource and also update that new version to setup.py of shopify_python_api (Or use git link as a dependency source).

After that this branch should be merged and that would also resolve our original issue with https://github.com/Shopify/shopify_python_api/pull/86 :)

Great job @gavinballard!

gavinballard:
Thanks for taking the time to double check this, @mentholi!

The ideal way to get all this merged in my mind is:

- Create a 2.1.1 release of `pyactiveresource`;
- I'll update this PR to depend on that new release;
- This branch should be green and mergeable;
- After that, the two other outstanding PRs should likewise be green and mergeable.

Does that sound reasonable, @kevinhughes27?

kevinhughes27:
:+1:

gavinballard:
:) Great! Just let me know when the new release is pushed and we can get these suckers merged.

kevinhughes27:
ping @gavinballard new version of pyactiveresource is now on PyPi

kevinhughes27:
ping me as the branches go green and I will merge them in

gavinballard:
@kevinhughes27 Ping! :)

PR #94, opened 2015-04-14
Closes #85: RecurringApplicationCharge.current() does not work as expected

### LINKED ISSUE #85: RecurringApplicationCharge.current() does not work as expected
In RecurringApplicationCharge class there is class method "current" that doesn't work as expexted. 

## Problem
I have few RecurringApplicationCharge instances for my test shop but none of them has status=active. I was expecting to get None as a result when using current() method but instead a RecurringApplicationCharge with status rejected was returned.

```python
class RecurringApplicationCharge(ShopifyResource):

    @classmethod
    def current(cls):
        return cls.find_first(status="active")
```

## Excpected result 
First RecurringApplicationCharge object with status=active should be returned or None if no matches found. 

Anyway I noticed that documentation (http://docs.shopify.com/api/recurringapplicationcharge) states that it is only possible to filter with field "since_id". If this is correct, then this method should be removed.

gavinballard:


kevinhughes27:
released.

gavinballard:
Thanks guys!

PR #96, opened 2015-05-05
Fix unicode buffer issue

gavinballard:
In some cases, the `params` passed to `validate_hmac` could be in Unicode format. This causes issues for the `hmac.compare_digest` function (see [this similar issue](https://github.com/mitsuhiko/werkzeug/issues/537)).

This issue would not be apparent to any system running Python < 2.7.7, as `compare_digest` isn't available on 2.7.6 and lower.

This PR adds a failing test for the Unicode test and fixes the issue by ensuring both arguments to `compare_digest` are UTF-8 encoded.

### DIFF
--- a/shopify/session.py
+++ b/shopify/session.py
@@ -99,8 +99,8 @@ class Session(object):
         if 'hmac' not in params:
             return False
 
-        hmac_calculated = cls.calculate_hmac(params)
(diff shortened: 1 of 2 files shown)


kevinhughes27:
thanks!

PR #98, opened 2015-05-12
Escape ampersand in params or equal in keys before signing with hmac.

dylanahsmith:
@kevinhughes27 for review

Porting https://github.com/Shopify/shopify_api/pull/188 to python.

I ended up using str.replace to escape the same subset of characters, and to make sure the case of the URI escape codes is the same.


kevinhughes27:
lgtm

Issue #100, opened 2015-05-21
Cannot create customer with password


sephii:
The following code will fail with an error `{u'password_confirmation': [u"doesn't match Password"]}`:

```python
import shopify

c = shopify.Customer()
c.email = 'foo@bar.com'
c.password = 'foo'
c.password_confirmation = 'foo'
c.save()
```

`c.__dict__` is:

```python
{'errors': <pyactiveresource.activeresource.Errors object at 0x431df10>, '_initialized': True, '_prefix_options': {}, 'klass': <class 'shopify.resources.customer.Customer'>, 'attributes': {'password_confirmation': 'foo', 'email': u'foo@bar.com'}, 'password': 'foo'}
```
As you can see the `password` attribute is not in the `attributes`. A workaround is to set the password in the `__init__`:

```python
c = shopify.Customer(attributes={'password': 'foo', 'password_confirmation': 'foo'})
```

dylanahsmith:
I wasn't able to reproduce the problem.  I tried writing a customer test for this case and ran it with python 2 and 3.  I also tried using a production console and didn't receive that problem, although I needed to increase the password length to avoid a validation error (minimum password length is 5).

What version of python, pyactiveresource and shopify_python_api are you using?

I'll create a pull request with the test I wrote and cc you on it.

dylanahsmith:
Closing since I couldn't reproduce.  But if you could modify the test that I added in https://github.com/Shopify/shopify_python_api/pull/105 to reproduce the issue, then we can get it fixed if this is an issue.

kaiserama:
I realize this is closed, but just FYI - I encountered the same problem as @sephii.  Using his workaround worked for me as well.

sarendsen:
FYI: this bug still exists.

PR #105, opened 2015-07-10
Add customer tests

dylanahsmith:
cc @sephii and @kevinhughes27 

I noticed issue #100 and that we are missing customer tests.  Although, I wasn't able to reproduce that specific issue, I thought we should still add those tests.


kevinhughes27:
:+1:

PR #110, opened 2015-08-18
Support for independently creating/modifying product images (closes #88)

### LINKED ISSUE #88: Support for independently creating/modifying product images.
Please correct me if I'm wrong, but at the moment the only way to create/update product images through the Python API is directly through a POST/PUT call to the product endpoint.

As per the [Shopify API documentation on product images](http://docs.shopify.com/api/product_image), we should be able to make POST and PUT calls directly to `/admin/products/#{id}/images.json` and `PUT /admin/products/#{id}/images/#{id}.json`. However, this doesn't seem to work currently with the API as the prefix for the `Image` class isn't properly being built. Thus, something like this...

```python
image = shopify.Image()
image.src = "http://example.com/example.png"
image.product_id = 123456789
image.save()
```

... tries to POST to `/admin/products//images.json` which returns a 406.

Looking at the source, I think a fix would be to implement the `prefix()` class method, similar to what's been done with the `Variant` class. Just wanted to run this by folks who might be more familiar with this to see if that might be a valid solution.

asiviero:
As @gavinballard points out, a small change in Image class makes it possible for it to behave just as the Variant class, so I added that code to it, tested on my dev store and wrote a test for it.

### DIFF
--- a/shopify/resources/image.py
+++ b/shopify/resources/image.py
@@ -8,6 +8,14 @@ import re
 class Image(ShopifyResource):
     _prefix_source = "/admin/products/$product_id/"
 
+    @classmethod
+    def _prefix(cls, options={}):
+        product_id = options.get("product_id")
+        if product_id:
+            return "/admin/products/%s" % (product_id)
+        else:
+            return "/admin"
+
     def __getattr__(self, name):
         if name in ["pico", "icon", "thumb", "small", "compact", "medium", "large", "grande", "original"]:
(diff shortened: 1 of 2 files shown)


kevinhughes27:
@awd for triage

asiviero:
ping?

kevinhughes27:
@joshubrown can you get someone on this

flux627:
Is there anything preventing this from being merged? I'd really like to use this on the master branch.

kevinhughes27:
I am ok with this.

PR #129, opened 2015-12-11
Adds suport to Shipping Zones (closes #126)

### LINKED ISSUE #126: Add support for the new Shipping Zone resource.
As per https://ecommerce.shopify.com/c/api-announcements/t/shipping-zones-api-and-changes-to-the-countries-api-307687.

asiviero:
Hey guys,

This was easier than I thought since most of the heavy lifting is done by pyactiveresource.

Did some basic testing with my dev account and data seems fine

### DIFF
--- a/shopify/resources/__init__.py
+++ b/shopify/resources/__init__.py
@@ -46,5 +46,6 @@ from .policy import Policy
 from .smart_collection import SmartCollection
 from .gift_card import GiftCard
 from .discount import Discount
+from .shipping_zone import ShippingZone
 
 from ..base import ShopifyResource
--- a/shopify/resources/shipping_zone.py
+++ b/shopify/resources/shipping_zone.py
@@ -1,0 +1,5 @@
+from ..base import ShopifyResource
+
+
+class ShippingZone(ShopifyResource):
+    pass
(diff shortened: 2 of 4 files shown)


kevinhughes27:
Its pretty magical eh! Can you add a couple of tests? should be able to modify one of the other test files

asiviero:
Does it make sense to have a test for this since shipping zones is just a get request, without being able to modify anything? I mean, I guess there's no test for countries either

kevinhughes27:
Country should probably have a test as well! Its just a nice sanity check that it is generating the right url internally

asiviero:
Let me work on it

kevinhughes27:
Thanks!

asiviero:
There you go, not sure if it's the right way of doing it but I tracked the function that should make the collection url and tested against it, let me know if it was the right way of doing it

kevinhughes27:
I was thinking something more like this: https://github.com/Shopify/shopify_python_api/blob/master/test/article_test.py#L12

asiviero:
I guess that's more like it

kevinhughes27:
Nice!

PR #133, opened 2016-04-11
Remove the Python3 limitation

douglas:
Hello !

I'm studying the Python and noticed that Python3 support was added in the 2.1.0 version.

Thanks =)


kevinhughes27:
Good catch!

PR #117, opened 2015-09-29
Fixes #114 - Missing PBR requirement in test environment

ifnull:

Issue #114, opened 2015-09-17
Missing PBR requirement in test environment.


gavinballard:
When try to run `python setup.py test`, I ran into the following error:

```
(shopify_python_api_2.7.6)172-10-32-209:shopify_python_api gavin$ python setup.py test
running test
Searching for mock>=1.0.1
Best match: mock 1.3.0
Processing mock-1.3.0-py2.7.egg

Using /projects/forks/shopify_python_api/.eggs/mock-1.3.0-py2.7.egg
Searching for funcsigs
Best match: funcsigs 0.4
Processing funcsigs-0.4-py2.7.egg

Using /projects/forks/shopify_python_api/.eggs/funcsigs-0.4-py2.7.egg
Searching for pbr>=0.11
Best match: pbr 1.8.0
Processing pbr-1.8.0-py2.7.egg

Using /projects/forks/shopify_python_api/.eggs/pbr-1.8.0-py2.7.egg
running egg_info
writing requirements to ShopifyAPI.egg-info/requires.txt
writing ShopifyAPI.egg-info/PKG-INFO
writing top-level names to ShopifyAPI.egg-info/top_level.txt
writing dependency_links to ShopifyAPI.egg-info/dependency_links.txt
Traceback (most recent call last):
  File "setup.py", line 44, in <module>
    'Topic :: Software Development :: Libraries :: Python Modules']
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/local/Cellar/python/2.7.10_2/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/Users/gavin/.pyenv/versions/shopify_python_api_2.7.6/lib/python2.7/site-packages/setuptools/command/test.py", line 142, in run
    self.with_project_on_sys_path(self.run_tests)
  File "/Users/gavin/.pyenv/versions/shopify_python_api_2.7.6/lib/python2.7/site-packages/setuptools/command/test.py", line 106, in with_project_on_sys_path
    self.run_command('egg_info')
[...]

PR #134, opened 2016-04-15
Add tox for testing and Python 3.5 test support

bennylope:
Creates isolated test environments for conveniently testing against multiple Python versions.

The tox configuration here tests against Python 2.7, Python 3.4, Python 3.5, PyPy, and PyPy3 by default. Given that the issues referenced in #114 and #117 have to do with build isolation, this should resolve those issues (have not experienced them, that much I'm certain about).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: python
 python:
   - "2.7"
   - "3.4"
+  - "3.5"
 
 # command to install dependencies
 install: "python setup.py install"
--- a/README.md
+++ b/README.md
@@ -253,6 +253,24 @@ To run tests, simply open up the project directory in a terminal and run:
 python setup.py test
 ```
 
+Alternatively, use [tox](http://tox.readthedocs.org/en/latest/) to
+sequentially test against different versions of Python in isolated
+environments:
+
+```shell
+pip install tox
+tox
+```
+
+See the tox documentation for help on running only specific environments
+at a time. The related tool [detox](https://pypi.python.org/pypi/detox)
+can be used to run tests in these environments in parallel:
+
+```shell
+pip install detox
+detox
+```
+
 ## Limitations
 
 Currently there is no support for:
--- a/setup.py
+++ b/setup.py
@@ -39,6 +39,7 @@ setup(name=NAME,
                    'Programming Language :: Python :: 2.7',
                    'Programming Language :: Python :: 3',
                    'Programming Language :: Python :: 3.4',
+                   'Programming Language :: Python :: 3.5',
                    'Topic :: Software Development',
                    'Topic :: Software Development :: Libraries',
                    'Topic :: Software Development :: Libraries :: Python Modules']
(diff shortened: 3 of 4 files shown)


kevinhughes27:
Can you give a bit more context here? Is this so that I can test against multiple python versions locally?

How would I actually use this locally?

How does this specifically fix issues with people's local setup as in the issues you described? Is tox a common utility in the python community (sorry haven't spent much time working with python the last few years)

I am a bit against adding more tooling here since it adds complexity and we're not python experts so simple is better!

bennylope:
> Is this so that I can test against multiple python versions locally?

Yes, that's the *primary* purpose of using tox. The way it does this is by creating and managing virtual environments for *each* version of Python (or each environment, which may include a matrix of environments, much like Travis).

> How would I actually use this locally?

You'd run the `tox` command. I added some [documentation to the README](https://github.com/Shopify/shopify_python_api/pull/134/files#diff-04c6e90faac2675aa89e2176d2eec7d8) to show how to use this.

I just removed `pypy` and `pypy3` from the default environment set, since that could cause confusion. When you run the `tox` command it will test against each specified environment. In this case it will look for a Python 2.7 interpreter, a Python 3.4 interpreter, and a Python 3.5 interpreter, build an isolated virtualenv for each Python version, and run the tests in that isolated virtualenv.

> How does this specifically fix issues with people's local setup as in the issues you described?

I'm not 100% sure it will, but relatively confident. It's creating isolated [virtualenvs](http://docs.python-guide.org/en/latest/dev/virtualenvs/), separate from whatever other Python virtualenv they're working with. 

> Is tox a common utility in the python community?

Yes. The best part about adding tox support is that _no one is required to use it_. It makes it very easy to test against multiple versions of Python right away so you don't make incompatible changes, but it works by running whatever test script you feed it. Here it's running `python setup.py test`.

> I am a bit against adding more tooling here since it adds complexity and we're not python experts so simple is better!

This, then, is the type of tooling you _do_ want to add! No one wants to manually test against every supported Python version. It's simple to set up, there's no requirement to use it, and it lets you test against Python versions without having to think about it.

bennylope:
Looking through the original issue again and my own local codebase, `pbr` is required by `mock`. A secondary change would be to update the mock version required, however I would strongly recommend adding tox support for any Python project supporting multiple versions of Python.

What other projects are using tox to make testing easier?

- [Flask](https://github.com/pallets/flask/blob/master/tox.ini)
- [Jinja2](https://github.com/pallets/jinja/blob/master/tox.ini)
- [Ansible](https://github.com/ansible/ansible/blob/devel/tox.ini)
- [speedtest-cli](https://github.com/sivel/speedtest-cli/blob/master/tox.ini)
- [Rackspace Cloud Files API](https://github.com/rackspace/pyrax/blob/master/tox.ini)
- [SQLAlchemy](https://bitbucket.org/zzzeek/sqlalchemy/src/35e93db7ffb5806df22e997fde1f966fa8c453b0/tox.ini?fileviewer=file-view-default)

And soon... shopify_python_api! :dancers:

kevinhughes27:
Thanks for the detailed explanation. Seems like it would be useful so I am okay with this! Thanks for putting this together and explaining it.

PR #136, opened 2016-04-18
Enable keyword arguments when requesting metafields

### LINKED ISSUE #135: metafields() method should be able to take options parameters for limit and page
```
import shopify
product = shopify.Product.find(5972485446)
metafields = product.metafields()
print(len(metafields))
> 50
metafields = product.metafields(limit=250)
> TypeError: metafields() got an unexpected keyword argument 'limit'
```
I looked into the code, and it seems like it may be a simple fix. If I come up with something I'll submit a pull request, if I can't I'll let it be known here so somebody else can try tackling it.

flux627:
Resolves #135


kevinhughes27:
Can you add some tests please!

flux627:
I added tests for the `metafields_count` method, even though it looks like no other resource tests for count functionality. I can remove it if you think it's not necessary.

test/fixtures/metafields_count.json
```diff
@@ -0,0 +1 @@
+{"count":2}
```
kevinhughes27: can you add the newline char here please

shopify/mixins.py
```diff
@@ -11,8 +11,15 @@ def count(cls, _options=None, **kwargs):
...
-    def metafields(self):
-        return shopify.resources.Metafield.find(resource=self.__class__.plural, resource_id=self.id)
+    def metafields(self, _options=None, **kwargs):
+        if _options is None:
+            _options = kwargs
+        return shopify.resources.Metafield.find(resource=self.__class__.plural, resource_id=self.id, **_options)
+
+    def metafields_count(self, _options=None, **kwargs):
+        if _options is None:
+            _options = kwargs
```
kevinhughes27: can you explain this `_options` code please

flux627: I took the pattern from the `count` method of the `Countable` class, seen just above in the same file. Basically, it allows you to either specify parameters with a dictionary or with keyword arguments. This is consistent with the behavior of the rest of the API.

kevinhughes27: okay cool that makes sense! Sorry I am just not super familiar with python or this repo anymore so figured I would ask.

flux627:
Anything else needed before we're good to merge?

kevinhughes27:
nope. Thanks for the PR!

PR #156, opened 2016-09-12
Removed all references in code to deprecated signature parameter

jamiemtdwyer:
Removing all references in the code to the deprecated `signature` parameter which is no longer provided by Shopify in the interest of cleaning up the codebase and reducing confusion for partners.

I removed the test for `test_return_token_if_hmac_is_valid_but_signature_also_provided` since its no longer applicable but I wasn't sure about `test_raise_error_if_params_passed_but_signature_omitted` - is this testing if `signature` and `hmac` are omitted?

@kevinhughes27 @dylanahsmith @ShayneP

### DIFF
--- a/shopify/session.py
+++ b/shopify/session.py
@@ -136,7 +136,7 @@ class Session(object):
         """
         def encoded_pairs(params):
             for k, v in six.iteritems(params):
-                if k not in ['signature', 'hmac']:
+                if k != 'hmac':
                     # escape delimiters to avoid tampering
                     k = str(k).replace("%", "%25").replace("=", "%3D")
                     v = str(v).replace("%", "%25")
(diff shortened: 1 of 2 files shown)


dylanahsmith:
:+1:

> test_raise_error_if_params_passed_but_signature_omitted - is this testing if signature and hmac are omitted?

It was basically making sure an attacker couldn't just remove the signature and have the params be considered valid.  We care about the same thing with the hmac signature.

kevinhughes27:
👍

jamiemtdwyer:
Do I need to do anything with pip or easyinstall?

kevinhughes27:
Nope iirc you can merge and push the new version using shipit

dylanahsmith:
It looks like the last attempts at releasing through shipit failed.  We have https://github.com/Shopify/shopify_python_api/blob/master/RELEASING documenting how to release manually if necessary.

jamiemtdwyer:
:ship:

PR #164, opened 2016-10-13
Improvements

paladinlogic:
Some basic improvements to related objects (eg adding images to products/variants and fulfillments to orders). Also, Variant was coming back with the wrong urls. Not sure if fix completely correct, but was working for us.

### DIFF
--- a/shopify/resources/order.py
+++ b/shopify/resources/order.py
@@ -24,12 +24,12 @@ class Order(ShopifyResource, mixins.Metafields, mixins.Events):
     def add_fulfillment(self, fulfillment):
         if not self.id:
             self.save()
-        fulfillment.order_id = self.id
+        fulfillment.attributes['order_id'] = self.id
 
         result = fulfillment.save()
         if result:
             if not isinstance(self.attributes.get('fulfillments'), Iterable):
-                self.fulfillments = []
-            self.fulfillments.append(fulfillment)
+                self.attributes['fulfillments'] = []
+            self.attributes['fulfillments'].append(fulfillment)
 
         return result
--- a/shopify/resources/order.py
+++ b/shopify/resources/order.py
@@ -19,3 +19,10 @@ class Order(ShopifyResource, mixins.Metafields, mixins.Events):
 
     def capture(self, amount=""):
         return Transaction.create({"amount": amount, "kind": "capture", "order_id": self.id})
+
+    def add_fulfillment(self, fulfillment):
+        if not self.id:
+            self.save()
+        fulfillment.order_id = self.id
+        fulfillment.save()
+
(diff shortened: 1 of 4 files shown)


jamiemtdwyer:
I'm personally not a huge fan of adding find by sku to this library , thoughts @dylanahsmith @kevinhughes27 ?

Everything seems fine to be, I'll do a thorough code review sometime next week.

dylanahsmith:
> I'm personally not a huge fan of adding find by sku to this library

I agree.  I don't think we have any uniqueness constraints on the sku, which is why he has to use _find_every to get the variant.  If there is another match, then it would just be ignored.

kevinhughes27:
Agreed and SKU is not unique 👍

paladinlogic:
Maybe I misread what the documentation meant by:

> A unique identifier for the product in the shop.

Is there a better option other than id?

shopify/resources/variant.py
```diff
@@ -8,12 +8,19 @@ class Variant(ShopifyResource, mixins.Metafields):
     @classmethod
     def _prefix(cls, options={}):
         product_id = options.get("product_id")
-        if product_id:
+        if options.get('is_new'):
```
dylanahsmith: Does this actually fix something?  This PR is very hard to review because of these scattered changes without much context.

paladinlogic:
Sorry, haven't had much time to devote on this project in the past few months. I'll have to review each of these and remember what was going on with them.

shopify/resources/order.py
```diff
@@ -19,3 +20,16 @@ def transactions(self):
 
     def capture(self, amount=""):
         return Transaction.create({"amount": amount, "kind": "capture", "order_id": self.id})
+
+    def add_fulfillment(self, fulfillment):
+        if not self.id:
+            self.save()
```
paladinlogic: That makes sense, if not at least for consistency. I'll submit the modification soon.

shopify/resources/order.py
```diff
@@ -19,3 +20,16 @@ def transactions(self):
...
+        fulfillment.attributes['order_id'] = self.id
+
+        result = fulfillment.save()
+        if result:
+            if not isinstance(self.attributes.get('fulfillments'), Iterable):
+                self.attributes['fulfillments'] = []
```
paladinlogic: I'm not sure I follow. If the order already has a fulfillment, it will already be an Iterable and this line will be skipped.

shopify/resources/variant.py
paladinlogic: IIRC, we were getting the incorrect URI for Options. This specific change is OBE for us anyway as our solution won't be using Options anymore anyway.

nwtn:
Closing because there’s been no activity for > 2 years

PR #177, opened 2017-02-21
fix #147: ensure image attachments are strings to be JSON-encodable

### LINKED ISSUE #147: Attaching Images - Decode required for Python 3 compatibility
When attaching an image from a local file:

```
with open("some_image.jpg","rb") as f:
    img = f.read()
    image1.attach_image(data=img)
```

img will be bytes while the attachment needs to be a string to the JSONified. So in image.py:

```
def attach_image(self, data, filename=None):
    self.attributes["attachment"] = base64.b64encode(data).decode() #decode() for Python 3
    if filename:
        self.attributes["filename"] = filename
```

I am not sure how to keep it Python 2 compatible.

ressy:
As mentioned in #147, `attach_image` from `shopify.image` under Python 3 creates an image object that can't be saved, because the bytes array for the attachment can't be JSON-encoded.  This decodes the bytes back into a plain string.  (In Python 2 this same change seems to work without any complications, since it takes a string object to start with and decodes it into a unicode string containing the same ASCII-only bytes it started with.)

Thanks @pntt.

### DIFF
--- a/shopify/resources/image.py
+++ b/shopify/resources/image.py
@@ -23,7 +23,7 @@ class Image(ShopifyResource):
             return super(Image, self).__getattr__(name)
 
     def attach_image(self, data, filename=None):
-        self.attributes["attachment"] = base64.b64encode(data)
+        self.attributes["attachment"] = base64.b64encode(data).decode()
         if filename:
             self.attributes["filename"] = filename
 
--- a/test/image_test.py
+++ b/test/image_test.py
@@ -1,5 +1,6 @@
 import shopify
 from test.test_helper import TestCase
+import base64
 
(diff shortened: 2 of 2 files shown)


jamiemtdwyer:
Hey ressy,

Thanks for this. Just to be clear, this is back-compatible with Python 3? I'll try to test and merge this by end-of-week and roll it up into the next major release.

ressy:
Sorry for the delayed reply. Yes, in my tests this fix works in both Python 2.7 and Python 3.5 (is that what you mean?). Thanks!

jamiemtdwyer:
Sorry for the delay folks – this should have been merged a long time ago.

PR #178, opened 2017-02-21
Add draft orders

lugray:
Add support for draft orders.


shopify/version.py
```diff
@@ -1 +1 @@
-VERSION = '2.1.8'
```
lugray: > MINOR version when you add functionality in a backwards-compatible manner, and

-semver.org

lugray:
I removed the version bump as we expect to add the complete action this week, so I'll bump the version once that's done.

PR #184, opened 2017-03-27
if the timestamp is missing, Dont Panic

orenmazor:
@kevinhughes27 @lugray

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ tags
 .DS_Store
 .eggs
 .python-version
+.cache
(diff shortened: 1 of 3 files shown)


jamiemtdwyer:
If timestamp isn't being returned in requests, I don't think that's an intentional change on Shopify's end... The timestamp is used by the client library to prevent replay attacks, and there hasn't been any similar changes to the Ruby library.

orenmazor:
@jamiemtdwyer yup. just making this a bit more resilient. turning a 500 into 400 (which is what it is… as the incoming request was malformed)

PR #186, opened 2017-03-30
Add support for customer invites

elsom25:
Adds `CustomerInvite` and exposes `Customer#send_invite`


kevinhughes27:
https://github.com/Shopify/shopify_python_api/blob/master/RELEASING

PR #187, opened 2017-04-03
Delete service.yml

SebastianSzturo:
Libraries should not have a `service.yml` AFAIK and partnerships-eng is not owning this.


kevinhughes27:
cc @maartenvg

PR #194, opened 2017-04-27
Add constant time fallback

EdOverflow:
See https://hackerone.com/reports/224096.


shopify/session.py
```diff
@@ -113,11 +113,15 @@ def validate_hmac(cls, params):
         hmac_to_verify = params['hmac'].encode('utf-8')
 
         # Try to use compare_digest() to reduce vulnerability to timing attacks.
-        # If it's not available, just fall back to regular string comparison.
         try:
             return hmac.compare_digest(hmac_calculated, hmac_to_verify)
         except AttributeError:
-            return hmac_calculated == hmac_to_verify
+            if len(hmac_calculated) != len(hmac_to_verify):
```
ivRodriguezCA: Can you switch the `print`s keywords for `return`s please. Thanks.

EdOverflow: Whoops.

EdOverflow:
I have now replaced `print` with `return`.

ivRodriguezCA:
Thank you @EdOverflow,

A member of our engineering team will look at this PR and merge it.

lugray:
This should include a version bump to 2.4.1 so that it can be released.

EdOverflow:
@lugray Does that look good to you? 🙂

shopify/version.py
```diff
@@ -1 +1 @@
-VERSION = '2.4.0'
+VERSION = '2.4.1'
```
lugray: This should have an associated changelog entry, and a `\n` at end of file.

lugray:
Also, can you add a test?

EdOverflow:
> Also, can you add a test?

Yes, I would be willing to add a test, but isn't there already one?

~~~python
def test_hmac_validation(self):
    # Test using the secret and parameter examples given in the Shopify API documentation.
    shopify.Session.secret='hush'
    params = {
      'shop': 'some-shop.myshopify.com',
      'code': 'a94a110d86d2452eb3e2af4cfb8a3828',
      'timestamp': '1337178173',
      'hmac': u('2cb1a277650a659f1b11e92a4a64275b128e037f2c3390e3c8fd2d8721dac9e2'),
    }
    self.assertTrue(shopify.Session.validate_hmac(params))
~~~

lugray:
You need a test that fails before your change, and passes after it.

EdOverflow:
OK, I see what you mean, but for there to be an error the test must be running on Python <2.7.7. Travis CI currently goes down to Python 2.7.12. @lugray Do you have any advice?

EdOverflow:
I could probably add the following to the `.travis.yml` file, but it may cause issues elsewhere:

~~~diff
python:
+  - "2.6"
   - "2.7"
   - "3.4"
   - "3.5"
~~~

lugray:
You could mock `compare_digest` to `raise AttributeError`

ofek:
A few things to note:

1. The logic of choosing which algorithm to use should occur at import time. Compare needs its own function.
2. This must not be in tests because this does not work in Python 3. Iteration of `bytes` in py3 yields an `int` unlike py2 so you can't/don't need to use `ord`.
3. In the case of different length byte strings this is not constant time. You should remove that and do `result = abs(len(hmac_calculated) - len(hmac_to_verify))`.

EdOverflow:
Thank you @ofek! What are your thoughts on this @ivRodriguezCA and @lugray?

lugray:
I disagree with the idea of not needing a test.  If we're making a change, particularly a security fix, there should be a regression test.  Thinking about this some more, this code is here only to support python < 2.7.  If we intend to support python < 2.7, then shouldn't we be testing against it?  I think you had the right idea when you suggested adding 2.6 to `.travis.yml`.  If it causes issues elsewhere, then we should fix those (because we're supporting 2.6). Or we should decide that we're not supporting 2.6, and that therefore this code isn't needed.  Thoughts?

clayton-shopify:
`compare_digest` was added to Python 2 in version 2.7.7 (May 31, 2014) and Python 3 in version 3.3 (September 29, 2012). So this change should be tested on < 2.7.7 and possibly also < 3.3.

That said, enough time has passed since `compare_digest` was added to Python that I'm not convinced this change is worthwhile. Ordinary string comparisons in Python are very fast (about 0.224 nanoseconds per byte on my machine), and so remote timing attacks are not practical.

EdOverflow:
> I disagree with the idea of not needing a test. If we're making a change, particularly a security fix, there should be a regression test. Thinking about this some more, this code is here only to support python < 2.7. If we intend to support python < 2.7, then shouldn't we be testing against it? I think you had the right idea when you suggested adding 2.6 to .travis.yml. If it causes issues elsewhere, then we should fix those (because we're supporting 2.6). Or we should decide that we're not supporting 2.6, and that therefore this code isn't needed. Thoughts?

@lugray You raised some interesting points and I hope your team can find a solution. Unfortunately, I am really busy at the moment and I do not whether I will have the time to finish updating this PR with tests. Would somebody be willing to take over from me or are you willing to wait a little while?

>  Ordinary string comparisons in Python are very fast (about 0.224 nanoseconds per byte on my machine), and so remote timing attacks are not practical.

@clayton-shopify I agree that this is not the easiest issue to exploit, but I am not entirely sure that I agree with your statement. I have seen CTFs where timing attacks were exploited in Python remotely. Also research has shown that it is possible to exploit this issue remotely:

> We find that LAN hosts can accurately resolve a 1 in the hundreds of nanoseconds, and that PlanetLab
hosts can resolve 1’s around 100µs, with the best hosts resolving 1 ≈ 30µs. [[1]]

[1]: http://www.cs.rice.edu/~dwallach/pub/crosby-timing2009.pdf

clayton-shopify:
> We find that LAN hosts can accurately resolve a 1 in the hundreds of nanoseconds, and that PlanetLab hosts can resolve 1’s around 100µs, with the best hosts resolving 1 ≈ 30µs. [1]

Python is well under that threshold, at about 0.0002 µs.

ofek:
@clayton-shopify Your math is a bit off there. 0.0002 µs == 0.2 ns which is impossible.

clayton-shopify:
> 0.2 ns which is impossible

I think it's correct. See p. 42 here: https://www.nccgroup.trust/globalassets/our-research/us/whitepapers/TimeTrial.pdf

EdOverflow:
@clayton-shopify Nanoseconds add up. The last character may require some simple brute forcing. ;)

Here is a quick and dirty PoC:

~~~python
import time, hmac


def timing(f):
    def wrap(*args):
        time1 = time.time()
        ret = f(*args)
        time2 = time.time()
        print '%s took %0.3f ms' % (f.func_name, (time2-time1)*1000.0)
        return ret
    return wrap

@timing
def timing_attack_diff_tokens():
    token_1 = "100000000000000000000000000000000"
    token_2 = "000000000000000000000000000000001"
    for i in range(200):
        if not token_1 == token_2:
            print i

@timing
def timing_attack_same_tokens():
    token_1 = "100000000000000000000000000000000"
    token_2 = "100000000000000000000000000000000"
    for i in range(200):
        if token_1 == token_2:
            print i

@timing
def constant_time_diff_tokens():
    token_1 = b"100000000000000000000000000000000"
    token_2 = b"000000000000000010000000000000000"
    for i in range(200):
        if not hmac.compare_digest(token_1, token_2):
            print i
        
@timing
def constant_time_same_tokens():
    token_1 = b"100000000000000000000000000000000"
    token_2 = b"100000000000000000000000000000000"
    for i in range(200):
        if hmac.compare_digest(token_1, token_2):
            print i

timing_attack_diff_tokens()
timing_attack_same_tokens()
constant_time_diff_tokens()
constant_time_same_tokens()
~~~

| Round   | timing_attack_diff_tokens | timing_attack_same_tokens | constant_time_diff_tokens | constant_time_same_tokens |
|---------|---------------------------|---------------------------|---------------------------|---------------------------|
| Round 1 | 2463 ms                   | 2365 ms                   | 2310 ms                   | 2329 ms                   |
| Round 2 | 2219 ms                   | 2175 ms                   | 2156 ms                   | 2188 ms                   |

ofek:
I acquiesce, but I'm quite skeptical of that paper's timings. I mean accessing L1 cache nowadays is still 0.5 ns minimum.

clayton-shopify:
> I acquiesce, but I'm still quite skeptical of that paper's timings. I mean accessing L1 cache nowadays is still 0.5 ns minimum.

Don't forget that comparisons are done eight bytes at a time on a 64-bit machine.

clayton-shopify:
@EdOverflow I don't think your methodology is correct because your sample size is very small and your timing measurements include unrelated and expensive operations such as allocating a `range` and calling `print`, which may also trigger garbage collections.

The following code demonstrates that the timing difference for 32-byte strings is about 2 nanoseconds:
```python
import time

N = 1000000000

token_1 = "000000000000000000000000000000000"
token_2 = "000000000000000000000000000000000"
token_3 = "100000000000000000000000000000000"

time1 = time.time()
for i in xrange(N):
    token_1 == token_2
time2 = time.time()
print str(10**9 * (time2 - time1) / N) + " ns per comparison for equal values"

time1 = time.time()
for i in xrange(N):
    token_1 == token_3
time2 = time.time()
print str(10**9 * (time2 - time1) / N) + " ns per comparison for unequal values"
```
That's about 0.1 nanoseconds per byte, which is comparable to Lawson & Taylor's 2010 results.

I agree that this is far from the cryptographic ideal, but it's also well below the amount that can be distinguished by practical timing attacks, so I think it would be reasonable to leave this code as-is. Python versions lacking `compare_digest` will soon be gone.

clayton-shopify:
Oops, I forgot to post the output:
```
83.3484480381 ns per comparison for equal values
81.529722929 ns per comparison for unequal values
```

EdOverflow:
RE:@clayton-shopify

> I don't think your methodology is correct because your sample size is very small and your timing measurements include unrelated and expensive operations such as allocating a range and calling print, which may also trigger garbage collections.

Good point.

> I agree that this is far from the cryptographic ideal, but it's also well below the amount that can be distinguished by practical timing attacks, so I think it would be reasonable to leave this code as-is. Python versions lacking compare_digest will soon be gone.

Fair enough, I will leave the current code as is. The only little thing I would suggest adding is a little comment above the block of code linking back to this discussion.

PR #195, opened 2017-04-27
Update RELEASING

jamiemtdwyer:


lugray:
Also needs instructions for creating `~/.pypirc` with a `[pypi]` section which must be created manually and without which the command will fail.

lugray:
👋  @jamiemtdwyer is this still something you're woking on?

jamiemtdwyer:
Good call – I'll add this in later today

jamiemtdwyer:
👀 @lugray 🙏

RELEASING
```diff
@@ -9,5 +9,7 @@ Releasing shopify_python_api
     git tag -m "Release X.Y.Z" vX.Y.Z
 6. Push the changes to github
     git push --tags origin master
-7. Upload the source package to pypi
-    python setup.py sdist upload
+7. Configure your credentials for pypi in $HOME/pypi.rc
+    (see https://docs.python.org/2/distutils/packageindex.html#the-pypirc-file for instructions)
```
lugray: Repository must be `https://pypi.python.org/pypi` which is different from the `https://www.python.org/pypi` default mentioned on that page.

jamiemtdwyer:
Some minor changes here, I think this is good to 🚢

PR #197, opened 2017-05-31
Price rules implementation

allanarmstrong:
Add [Price Rules API](https://help.shopify.com/api/price-rules)

### DIFF
--- a/shopify/resources/discount_code.py
+++ b/shopify/resources/discount_code.py
@@ -1,0 +1,8 @@
+from ..base import ShopifyResource
+
+
+'''
+Discount Code for Price Rules API
+'''
+class DiscountCode(ShopifyResource):
+    _prefix_source = "/admin/price_rules/$price_rule_id/"
--- a/shopify/resources/discount_code.py
+++ b/shopify/resources/discount_code.py
@@ -1,8 +1,4 @@
 from ..base import ShopifyResource
 
-
-'''
-Discount Code for Price Rules API
-'''
 class DiscountCode(ShopifyResource):
     _prefix_source = "/admin/price_rules/$price_rule_id/"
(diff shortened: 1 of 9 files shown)


shopify/resources/price_rule.py
```diff
@@ -0,0 +1,10 @@
+from ..base import ShopifyResource
+from .discount_code import DiscountCode
+
+class PriceRule(ShopifyResource):
+    def add_discount(self, discount_code = DiscountCode()):
```
felix-d: I'd probably be more explicit here and use restful verbs

`create_discount_code`

shopify/resources/discount_code.py
```diff
@@ -0,0 +1,8 @@
+from ..base import ShopifyResource
+
+
+'''
+Discount Code for Price Rules API
+'''
```
mlhamel: i would move this as a docstring of the `DiscountCode` class and use `"` instead of `'` ( 🚲 🏠 )

shopify/resources/price_rule.py
felix-d: Also here the discount_code will always be the same instance since the default param is evaluated once

```python
class Bar:
  pass

def test(foo = Bar()):
  print(foo)
  
test()  # <__builtin__.Bar instance at 0x7f74ecb1f170>
test()  # <__builtin__.Bar instance at 0x7f74ecb1f170>
```

shopify/resources/discount_code.py
felix-d: no other resource with a prefix use docstrings, I'd 🔥  it 🚲 🏠

shopify/resources/price_rule.py
jamiemtdwyer: agreed that `add_discount_code` is preferable here

allanarmstrong: I'm not sure about this one either. A similar default param is used in draft_order.py in the send_draft_order_invoice method.  Is no default preferable in this case or is there a better way to do it?

jamiemtdwyer:
I think this is ready to 🚢 – unless anyone has some last minute comments, I'll merge and deploy this tomorrow.

PR #203, opened 2017-07-07
fix singular name for OrderRisk

daleevans:
the API actually uses "risk" for the singular name, not "order_risk" which is what is auto-generated from the OrderRisk class name


jamiemtdwyer:
Sorry, we must have missed this – this was fixed in https://github.com/Shopify/shopify_python_api/pull/211.

We appreciate your contributions as always!

PR #205, opened 2017-07-26
Return scopes when requesting access token

saturn4er:
After user installed application, and we requesting an access token, it would be nice to also retrieve the scopes, which user gives to the application.


jamiemtdwyer:
I think this changes the API too drastically. There is an [AccessScope](https://help.shopify.com/en/api/reference/access/accessscope) introspection in the Admin API now, so it's probably better to use this. I'll add support for this in 4.0.0.

PR #211, opened 2017-10-26
Fix order risk resource name

ryan-ormsby:
Singular resource name for OrderRisk was previously `order_risk`, as determined by the resource name.

This change updates the resource name to be `risk` (as per the API specification).

Also bumps version to `2.5.1` for release.


jamiemtdwyer:
👍 🐑

jamiemtdwyer:
:shipit:

PR #216, opened 2017-11-21
Statsd support

orenmazor:
valuable for us. not sure if anybody else cares.

@kevinhughes27 @richardblair @perobertson

### DIFF
--- a/shopify/base.py
+++ b/shopify/base.py
@@ -19,12 +19,26 @@ class ShopifyConnection(pyactiveresource.connection.Connection):
 
     def _open(self, *args, **kwargs):
         self.response = None
+
         try:
             self.response = super(ShopifyConnection, self)._open(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


kevinhughes27:
Code looks good to me and is a good example of how to add telemetry to this package. I'm not sure how I feel about merging as is though because it adds an extra dependency that not everyone will use. Any ideas on how to make this more optional?

perobertson:
around the statsd code you can do something like this:
```python
try:
  from datadog import statsd
  tags = []
  statsd.increment('thing', tags=tags)
except ImportError:
  # datadog is optional
  pass
```

jamiemtdwyer:
If you write a short guide on how to do this, I'd be happy to add it to the README but I'll echo Kevin's statement on not adding dependencies that not every consumer will need.

wowkin2:
I don't think that this code should go to this repository, you can always patch library with anything you need, but not all developers use datadog.

PR #226, opened 2018-01-26
Remove deprecated Discount resource

jamiemtdwyer:
Remove deprecated Discount resource (previously mapped to the now-deprecated Discount API).


jordanliddle:
@alanly this is OK to 🔥,  yeah?

Issue #188, opened 2017-04-05
Add Support for Resource Feedback API


pbojinov:
Creating shop feedback and listing shop feedback

https://help.shopify.com/api/resource-feedback

jamiemtdwyer:
I'll get to work on this as soon as I have the cycles!

Issue #185, opened 2017-03-30
Add Support for Storefront Access Token


pbojinov:
Add support for creating, getting and deleting Storefront Access Tokens.

Sanjeev-Gupta:
Hi,

For public app, I am getting null when calling storefront api of access token. Can anyone tell me how will i create access token for storefront for public app.

jamiemtdwyer:
Closed by https://github.com/Shopify/shopify_python_api/pull/227

PR #227, opened 2018-01-30
Add StorefrontAccessToken and ResourceFeedback

jamiemtdwyer:
Adds two new resources to the package:
* StorefrontAccessToken
* ResourceFeedback

This PR brings this package to parity (at least in terms of resource inclusion) with the current state of the proper Shopify Admin API.

One minor caveat is that ResourceFeedback will not raise an exception on save for persisted resources ([as it does in the Ruby version of this package](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/resource_feedback.rb#L14-L17)), as `pyactiveresource` would need to be extended to support the `persisted` attribute on resources.

Closes https://github.com/Shopify/shopify_python_api/issues/188 & https://github.com/Shopify/shopify_python_api/issues/185.

### DIFF
--- a/shopify/resources/__init__.py
+++ b/shopify/resources/__init__.py
@@ -58,5 +58,7 @@ from .discount_code import DiscountCode
 from .marketing_event import MarketingEvent
 from .collection_listing import CollectionListing
 from .product_listing import ProductListing
+from .resource_feedback import ResourceFeedback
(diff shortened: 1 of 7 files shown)


jamiemtdwyer:
🐑

PR #239, opened 2018-03-14
Handle bytes data for asset attachments

erm:
After upgrading a project to Python 3, attempting to call `attach()` on an `Asset` object would result in the following error: `TypeError: Object of type 'bytes' is not JSON serializable`.

Using the decode method resolves this.


jamiemtdwyer:
Hey @erm, thanks for caching this. We had previously merged a fix to handle attaching images to products but not assets.

PR #244, opened 2018-05-11
Inventory API

herrri:
https://help.shopify.com/api/reference/inventory

Implemented:

* Inventory items
* Inventory levels

Location endpoint is already implemented. 

Includes API changes to handle the active resources and tests.

### DIFF
--- a/shopify/resources/inventory_level.py
+++ b/shopify/resources/inventory_level.py
@@ -1,0 +1,40 @@
+from ..base import ShopifyResource
+import shopify
+import json
+
+
+class InventoryLevel(ShopifyResource):
+
+    def __repr__(self):
+        return '%s(%s)' % (self._singular, self.available)
+
+    @classmethod
+    def adjust(cls, location_id, inventory_item_id, available_adjustment):
+        body = {
+            'inventory_item_id': inventory_item_id,
+            'location_id': location_id,
+            'available_adjustment': available_adjustment
+        }
+        resource = cls.post('adjust', body=json.dumps(body))
+        return InventoryLevel(InventoryLevel.format.decode(resource.body))
+
+    @classmethod
+    def connect(cls, location_id, inventory_item_id, relocate_if_necessary=False, **kwargs):
+        body = {
+            'inventory_item_id': inventory_item_id,
+            'location_id': location_id,
+            'relocate_if_necessary': relocate_if_necessary,
+        }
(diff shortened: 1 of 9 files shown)


jamiemtdwyer:
Hey @herrri, thanks for your contribution! I'm going to spend some time later today determining which PR should merge in and how to best handle this. Cheers!

shopify/resources/inventory_level.py
```diff
@@ -0,0 +1,40 @@
+from ..base import ShopifyResource
+import shopify
+import json
+
+
+class InventoryLevel(ShopifyResource):
+
+    def __repr__(self):
```
jamiemtdwyer: I believe this should overload `_element_path` rather than `__repr__`

test/inventory_level_test.py
```diff
@@ -0,0 +1,49 @@
+import shopify
+from test.test_helper import TestCase
+
+
+class InventoryLevelTest(TestCase):
```
jamiemtdwyer: Should probably include a test for `destroy` as well

jamiemtdwyer:
Some minor comments, but I've confirmed that this works by testing on my live shop.

Some nitpicky issues I have with the current state of the PR are in style and tests – since this is a versioned package, I don't see any harm in merging this in now and following up with those changes in a separate PR.

The only major issue here is that it does not seem to be possible to destroy an inventory level in any way.  I'll be sure to address that in the follow-up PR.

shopify/resources/inventory_level.py
```diff
@@ -0,0 +1,40 @@
...
+        return '%s(%s)' % (self._singular, self.available)
+
+    @classmethod
+    def adjust(cls, location_id, inventory_item_id, available_adjustment):
+        body = {
+            'inventory_item_id': inventory_item_id,
+            'location_id': location_id,
+            'available_adjustment': available_adjustment
+        }
+        resource = cls.post('adjust', body=json.dumps(body))
```
rnovacek: The `body` should be encoded, this throws an error:
```
TypeError: POST data should be bytes, an iterable of bytes, or a file object. It cannot be of type str.
```

Changing this line to:
```python
resource = cls.post('adjust', body=json.dumps(body).encode())
```
fixes the error for me.

jamiemtdwyer: Good catch. I'll fix this in #245

PR #245, opened 2018-05-12
Add improvements for InventoryLevel resource

jamiemtdwyer:
This PR also adds the following improvements to the InventoryLevel resource which was added in #244:
* style improvements (added trailing newlines where they were missing)
* overload `InventoryLevel._element_path` to point to correct location on server
* add support for destroying instances of InventoryLevel and test
* improves tests
* always encode request body in POST requests
* overload `InventoryLevel.__repr__` to provide a sane representation of an inventory level resource (include `location_id` and `inventory_item_id`)


jamiemtdwyer:
Cc @herrri, feel free to add yourself as a reviewer.

shopify/resources/inventory_level.py
```diff
@@ -38,3 +43,10 @@ def set(cls, location_id, inventory_item_id, available, disconnect_if_necessary=
         }
         resource = cls.post('set', body=json.dumps(body))
         return InventoryLevel(InventoryLevel.format.decode(resource.body))
+
+    def is_new(self):
+        return False
+
+    def destroy(self):
+        options = {"inventory_item_id": self.inventory_item_id, "location_id": self.location_id}
+        return self.__class__.connection.delete(self._element_path(query_options=options), self.__class__.headers)
```
mlhamel: just calling `self.connection...` and `self.headers...` just should just work, self is an instance of the `__class_` then it should have the same attributes.

herrri: Looks good @mlhamel . Didn't have time to take a look at this before now, but had something similar going on.

shopify/resources/inventory_level.py
```diff
@@ -5,8 +5,13 @@
...
 
-    def __repr__(self):
-        return '%s(%s)' % (self._singular, self.available)
+    @classmethod
+    def _element_path(cls, prefix_options={}, query_options=None):
+        if query_options is None:
+            prefix_options, query_options = cls._split_options(prefix_options)
+
+        return "%s%s.%s%s" % (cls._prefix(prefix_options)+'/', cls.plural,
+                              cls.format.extension, cls._query_string(query_options))
```
herrri: Is there a way to have more informative way for the object representation than this `inventory_level(None)`, the inventory does not have it's own id? Clearly overriding the `__repr__` seems wrong approach.

jamiemtdwyer:
> Looks great, the only thing i'm wondering would it not make more more sense to split this into 2 different PRs the first one that would make the different fixes to the code and the second one that would bump the version and upgrade the CHANGELOG?

Done – this PR will simply add the improvements for the Inventory API and I'll release the new version in a separate PR.

jamiemtdwyer:
🚢

shopify/resources/inventory_level.py
jamiemtdwyer: FYI, this is where I landed

>>> x = shopify.InventoryLevel.find(inventory_item_ids=16428756291, location_ids=13968638008)
>>> x
[inventory_level(inventory_item_id=16428756291, location_id=13968638008)]

PR #262, opened 2018-09-09
Fix HMAC when query param is a list

Maks3w:


jamiemtdwyer:
Hey @Maks3w, this looks good to me – IIRC this can happen if a merchant logs into the application from an "App Link" action in the admin with multiple resources selected.

Would you mind writing a test case in `session_test.py`? I'd like to be able to get this into the 4.0.0 release. Thanks!

jamiemtdwyer:
Amazing, thanks for your contribution!

PR #264, opened 2018-09-20
[WIP] Add support for Checkout API

jamiemtdwyer:
Adds support for the [Checkout API](https://help.shopify.com/en/api/reference/sales_channels/checkout).

Support for the new Checkout API has been added under the `Checkout` resource. The previous `Checkout` resource, which corresponds to the [Abandoned Checkout API](https://help.shopify.com/en/api/reference/orders/abandoned_checkouts) has been moved to the `AbandonedCheckout` resource.

Release version will be bumped to 4.0.0 as per [semver](https://semver.org/), since the resulting API is backwards incompatible. I will bump the version number in a later PR – there are some other small changes I want to make before bumping the major version number again.


jamiemtdwyer:
There seems to be a bug that's causing the extra headers in the `_headers` attribute to not be sent to the server.

I'm not sure if this is a bug in this library or pyactiveresource. The README for pyactiveresource implies that my usage here is correct, so it might be due to some custom logic in this library. Hopefully it doesn't take me too long to track down the issue.

jamiemtdwyer:
This is much more difficult than I anticipated. It's likely that I'll end up releasing the other changes in an interim minor version patch version and hold off on this one for a while.

jamiemtdwyer:
Adding the [WIP] label back to this PR. Do not merge.

9at8:
I tried it on my local setup and the headers are only present for the `Checkout` resource. (tried `Checkout` and `Product`) Maybe there's something wrong with the tests?

manumilou:
Some kind of race condition when the tests run ? The accessor `get_headers` does not seem threadsafe anymore.

nwtn:
@ryan-ormsby any thoughts about the current state of this?

ryan-ormsby:
@nwtn Haven't looked at any of the changes past the initial PR when @jamiemtdwyer ran into issues, so not sure current state.

andyw8:
Closing as there hasn't been activity on this for a long time.

PR #161, opened 2016-10-06
Limits

paladinlogic:
Port of the Limits submodule from the Ruby api library.

PR #266, opened 2018-09-21
Limits

jamiemtdwyer:
Successor to #161.

To be honest, I'm sure if this is the best approach. I've added a `Limits` class which can be accessed through `shopify.Limits` but I'm not sure how "pythonic"  this approach is. Is there some way to make the methods available directly on the `shopify` instance? Would this be preferable?

Tests were a bit of a pain for me. If anyone is more familiar with `mock`, feel free to correct my usage here.

### DIFF
--- a/shopify/limits.py
+++ b/shopify/limits.py
@@ -1,0 +1,31 @@
+import shopify
+
+class Limits(object):
+    CREDIT_LIMIT_HEADER_PARAM = 'x-shopify-shop-api-call-limit'
+
+    @classmethod
+    def credit_left(cls):
+        return cls.credit_limit() - cls.credit_used()
+
+    @classmethod
+    def is_credit_maxed(cls):
+        return cls.credit_left() <= 0
+
+    @classmethod
+    def credit_limit(cls):
+        return int(cls._api_credit_limit_param()[1])
+
+    @classmethod
+    def credit_used(cls):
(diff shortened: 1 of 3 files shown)


shopify/limits.py
```diff
@@ -0,0 +1,39 @@
...
+    def credit_left(cls):
+        return cls.credit_limit() - cls.credit_used()
+
+    available_calls = credit_left
+
+    @classmethod
+    def is_credit_maxed(cls):
+        return cls.credit_left() <= 0
+
+    is_maxed = is_credit_maxed
```
cfournie: You could make this a read-only property instead of a function like in `base.py`, in which case you can use [this syntax](https://github.com/Shopify/shopify_python_api/blob/3fc6c0187b8cf37f527f56b2a8b8c71eab8d8840/shopify/base.py#L57-L58) here, e.g.
```python
    is_maxed = property(is_credit_maxed, None, None, "Docstring")
```

jamiemtdwyer: I think I may just remove the aliasing all together. It doesn't feel very "pythonic" to me.

jamiemtdwyer:
I decided to get rid of the aliasing because it seemed messy to me. The only reason I did it in the first place was to make it more consistent with the Ruby gem, but I really don't think it's necessary. I think the provided method names are clear enough.

PR #270, opened 2018-09-28
Add inventory_levels method for Location resource

jamiemtdwyer:
Add support for non-CRUD method `inventory_levels` on Location resource.


jamiemtdwyer:
Nice catch, mind taking another look?

PR #276, opened 2018-10-16
Add TenderTransaction resource

krispenney:
**Issue**: https://github.com/Shopify/money-transactions/issues/688

As we are preparing to launch the TenderTransaction api, we will need to make it available for use in our SDKs. This PR adds a new TenderTransaction resource as well as a json fixture.

This PR adds a TenderTransaction resource to the library, as well as a json fixture to be used for testing.

### DIFF
--- a/shopify/resources/tender_transaction.py
+++ b/shopify/resources/tender_transaction.py
@@ -1,0 +1,4 @@
+from ..base import ShopifyResource
+
+class TenderTransaction(ShopifyResource):
+    _prefix_source = "/admin/"
--- a/shopify/resources/tender_transaction.py
+++ b/shopify/resources/tender_transaction.py
@@ -1,4 +1,4 @@
 from ..base import ShopifyResource
 
 class TenderTransaction(ShopifyResource):
-    _prefix_source = "/admin/"
+    pass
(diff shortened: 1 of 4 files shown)


shopify/resources/tender_transaction.py
```diff
@@ -0,0 +1,4 @@
+from ..base import ShopifyResource
+
+class TenderTransaction(ShopifyResource):
+    _prefix_source = "/admin/"
```
jamiemtdwyer: This line is superfluous since there are no prefix options for this resource

test/tender_transaction_test.py
```diff
@@ -0,0 +1,12 @@
+import shopify
+from test.test_helper import TestCase
+
+class TenderTransactionTest(TestCase):
```
jamiemtdwyer: can we add a test for the `#show` action as well? Generally I prefer for the tests to cover all of the available controller actions so they act as documentation for the package

krispenney: Talked about this on slack, there's currently only an `index` route on the controller. In the future, we could add scoping on orders if a use-case appears.

shopify/resources/tender_transaction.py
krispenney: 👍 I've removed this

PR #277, opened 2018-10-17
Added currency endpoint

grcooper:
Adding this endpoint following similar examples as transactions. Docs being updated are here: https://github.com/Shopify/shopify/pull/174966


jamiemtdwyer:
Don't worry about releasing a new version. I have a few more changes that I want to roll up into the next release.

PR #287, opened 2018-12-11
Python 3 compatibility fix

Maks3w:


Maks3w:
Corporate CLA has be signed.

PR #292, opened 2019-02-24
Some Improvements, fixed API Call Limits

antwal:
I have fixed some things, the important parts that you have to check are good for you are:

The version change with release, to support in the release name the levels "ALPHA, BETA, RELEASE_CANDIDATE, FINAL" as specified:

```
# version_info format: (MAJOR, MINOR, MICRO, RELEASE_LEVEL, SERIAL)
# inspired by Python's own sys.version_info, in order to be
# properly comparable using normal operarors, for example:
#  (6,1,0,'beta',0) < (6,1,0,'candidate',1) < (6,1,0,'candidate',2)
#  (6,1,0,'candidate',2) < (6,1,0,'final',0) < (6,1,2,'final',0)
```

I moved api call limits within the resource directory, but it has to be fixed
I changed the travis settings for the build

For the rest I have updated the changelog in the case check everything before accepting the PR.

PR #298, opened 2019-04-04
Fix tests

tylerball:
This PR splits some code off from @antwal larger PR #292  (thanks!) in order to get tests passing in master.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,27 @@
 language: python
 
-python:
-  - "2.7"
-  - "3.4"
-  - "3.5"
-  - "3.6"
-
 # command to install dependencies
-install: "python setup.py install"
+install:
+    - "python setup.py install"
 
 # command to run tests
-script: 'python setup.py test'
+script:
+    - python -m pytest -v
 
-notifications:
-  flowdock:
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,18 +1,27 @@
...
-
 # command to install dependencies
-install: "python setup.py install"
+install:
+    - "python setup.py install"
 
 # command to run tests
-script: 'python setup.py test'
+script:
+    - python -m pytest -v
```
nwtn: running this required me to `pip install mock` before it would work. which is fine, assuming that’s available inside travis; if not, maybe we could include it in that setup script.

also worth mentioning that `python setup.py test` worked for me out-of-the-box on this branch

shopify/limits.py
```diff
@@ -1,31 +1,61 @@
...
+    https://help.shopify.com/en/api/getting-started/api-call-limit
 
-    @classmethod
-    def credit_left(cls):
-        return cls.credit_limit() - cls.credit_used()
+    Conversion of lib/shopify_api/limits.rb
+    """
+    # num_requests_executed/max_requests
+    # Eg: 1/40
+    CREDIT_LIMIT_HEADER_PARAM = 'http_x_shopify_shop_api_call_limit'
```
nwtn: i’m confused why this switches to underscores from dashes

tylerball: You're right, I updated this to be exactly like the ruby implementation: https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/limits.rb#L14

.travis.yml
tylerball: Hmm, `python setup.py install` should get everything necessary, I'll look into this later.

tylerball:
It's GREEEEN

shopify/limits.py
jamiemtdwyer: In Ruby land, Rack transforms the headers to underscores

PR #300, opened 2019-04-09
Add API Versioning

tylerball:
TODO:
- [x] Merge #299 
- [x] Add docs/info to README

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,40 @@ The data itself is sent as XML over HTTP to communicate with Shopify,
 which provides a web service that follows the REST principles as
 much as possible.
 
+## !! Breaking change notice for version 5.0.0 !!
+
+### Changes to ShopifyAPI::Session
+Session creation requires a `version` to be set, see [Api Versioning at Shopify](https://help.shopify.com/en/api/versioning).
+
+To upgrade your use of ShopifyAPI you will need to make the following changes.
+
+```python
+shopify.Session(domain, token)
+```
+is now
+```python
+shopify.Session(domain, version, token)
+```
+
+The `version` argument takes a string for any known version and correctly coerce it to a `shopify.ApiVersion`.  You can find the currently defined versions [here](https://github.com/Shopify/shopify_python_api/blob/master/shopify/api_version.py#L27).
+
+For example if you want to use the `2019-04` version you would create a session like this:
+```python
+session = shopify.Session(domain, '2019-04', token)
+```
+if you want to use the `unstable` version you would create a session like this:
+```python
+session = shopify.Session(domain, 'unstable', token)
+```
+
+### URLs that have not changed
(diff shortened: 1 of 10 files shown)


shopify/session.py
```diff
@@ -28,26 +29,29 @@ def setup(cls, **kwargs):
...
+            original_version)
 
-        session = Session(domain, token)
+        session = Session(domain, token, version)
         shopify.ShopifyResource.activate_session(session)
         yield
         shopify.ShopifyResource.activate_session(original_session)
 
-    def __init__(self, shop_url, token=None, params=None):
+    def __init__(self, shop_url, token=None, version=None):
```
tylerball: Reading through the docs again, I think the order of these args should change

shopify/api_version.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @property
+    def name(self):
+        return self._name
+
+    def api_path(self, site):
+        return site + self._path
+
+    def __eq__(self, other):
+        return self.numeric_version == int(other.numeric_version)
```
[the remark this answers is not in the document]
tylerball: Went ahead and removed the version comparison as it's not being used.

README.md
```diff
@@ -13,6 +13,40 @@ The data itself is sent as XML over HTTP to communicate with Shopify,
...
+```python
+session = shopify.Session(domain, 'unstable', token)
+```
+
+### URLs that have not changed
+
+- OAuth URLs for `authorize`, getting the `access_token` from a code, and using a `refresh_token` have _not_ changed.
+  - get: `/admin/oauth/authorize`
+  - post: `/admin/oauth/access_token`
+  - get: `/admin/oauth/access_scopes.json`
```
nwtn: 🔥 .json

shopify/session.py
```diff
@@ -28,26 +29,28 @@ def setup(cls, **kwargs):
...
+        original_session = shopify.Session(original_domain, original_version, original_token)
 
-        session = Session(domain, token)
+        session = Session(domain, version, token)
         shopify.ShopifyResource.activate_session(session)
         yield
         shopify.ShopifyResource.activate_session(original_session)
 
-    def __init__(self, shop_url, token=None, params=None):
+    def __init__(self, shop_url, version=None, token=None):
```
nwtn: i think it’s fine because it’s not being used, but just calling out that you lost `params` here, in case that was an error

README.md
nwtn: ignore me, it’s a real resource. it should have `.json`

chrisbutcher:
Legit CI fails?

shopify/session.py
tylerball: Yep, we weren't doing anything with it so I :fire:

PR #301, opened 2019-04-09
release 5.0.0, define 2019-04 api version

tylerball:


Maks3w:
May CHANGELOG should be updated

PR #330, opened 2019-06-25
Fix setting site with basic auth URL

### LINKED ISSUE #314: Shopify API url not working
Hi,

Currently, I'm just playing around, with this library. I created a development store and a private app.
When I'm trying to create a product I got the following error:

Python 3.7.2
ShopifyAPI==5.0.1

  File "C:\Users\danish\AppData\Local\Programs\Python\Python37\lib\http\client.py", line 877, in _get_hostport
    port = int(host[i+1:])
ValueError: invalid literal for int() with base 10: '16xxxxxxx9exxxx5xxxxxxxxxxxxc5@my-chandola.myshopify.com'

Any help will be appreciated.

Thanks

tylerball:
Fixes #314 

Currently in the README we tell users that are using this library with private app credentials (basic auth) to use the `set_site` method to set their session URL. This method blindly sets the `site` class attribute and is referenced by some resource definitions like [Shop](https://github.com/Shopify/shopify_python_api/blob/master/shopify/resources/shop.py#L10), which results in incorrect URLs being requested. Ideally resources shouldn't have to care about the FQDN/site url when generating code for API Requests, but that's out of scope of this fix.

This fix strips out the `username:password` fragment from the site URL like [the underlying pyactiveresource method](https://github.com/Shopify/pyactiveresource/blob/master/pyactiveresource/connection.py#L218-L233) but retains the path, which we need to keep because its our reference to the API version. Now `ShopifyResource.site` `connection.site` and threaded sites are all equivalent.

### DIFF
--- a/test/session_test.py
+++ b/test/session_test.py
@@ -84,7 +84,7 @@ class SessionTest(TestCase):
             assigned_site = shopify.ShopifyResource.site
 
         self.assertEqual('https://testshop.myshopify.com/admin/api/unstable', assigned_site)
-        self.assertEqual('https://None/admin/api/unstable', shopify.ShopifyResource.site)
+        self.assertEqual('https://none/admin/api/unstable', shopify.ShopifyResource.site)
 
     def test_create_permission_url_returns_correct_url_with_single_scope_and_redirect_uri(self):
         shopify.Session.setup(api_key="My_test_key", secret="My test secret")
--- a/README.md
+++ b/README.md
@@ -85,15 +85,17 @@ service. pyactiveresource has to be configured with a fully authorized
 URL of a particular store first. To obtain that URL you can follow
 these steps:
 
-1.  First create a new application in either the partners admin or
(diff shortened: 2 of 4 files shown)


test/session_test.py
```diff
@@ -84,7 +84,7 @@ def test_temp_works_without_currently_active_session(self):
             assigned_site = shopify.ShopifyResource.site
 
         self.assertEqual('https://testshop.myshopify.com/admin/api/unstable', assigned_site)
-        self.assertEqual('https://None/admin/api/unstable', shopify.ShopifyResource.site)
+        self.assertEqual('https://none/admin/api/unstable', shopify.ShopifyResource.site)
```
tanema: I am guessing this is downcased now because the parsing of the host downcases the host?

tylerball: Yeah, I'm not exactly sure what usecase this particular test is after, but I imagine the case change doesn't matter because this isn't a real URL anyway.

PR #350, opened 2019-12-09
enable shopify_api.py to load local modules

slippers:
After activating the console via shopify_api.py i could not access local python modules.  I needed to add the current working directory to the sys paths before starting the console.


tanema:
Thanks for the change!

PR #352, opened 2020-01-20
Cursor Pagination

tanema:
### Problem
This library still doesnt support cursor based pagination. This will add in that functionality.

### Solution
This is based largely off of @emdemir work however I have made some critical changes that we have needed when doing this work in other repos

- Using the links directly to fetch the next page. This makes it a lot easier to manage 
- Exposing next and previous links so that they can be fetched across requests, if they are stored in a session.
- has_next and has_previous methods for introspecting if the data is available.
 
#### misc changes
- deployment fixed
- update pyactiveresource
- fixed an order test which was broken for some reason

### DIFF
--- a/shopify/collection.py
+++ b/shopify/collection.py
@@ -39,17 +39,17 @@ class PaginatedCollection(Collection):
         self._current_iter = None
         self._no_iter_next = kwargs.pop("no_iter_next", False)
 
-    def has_previous(self):
+    def has_previous_page(self):
         """Returns true if the current page has any previous pages before it.
         """
         return bool(self.previous_page_url)
 
-    def has_next(self):
+    def has_next_page(self):
         """Returns true if the current page has any pages beyond the current position.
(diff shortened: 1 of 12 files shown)


shopify/collection.py
```diff
@@ -0,0 +1,143 @@
...
+        """Returns true if the current page has any previous pages before it.
+        """
+        return bool(self.previous_page_url)
+
+    def has_next(self):
+        """Returns true if the current page has any pages beyond the current position.
+        """
+        return bool(self.next_page_url)
+
+    def previous(self, no_cache=False):
```
garethson: Nitpick: Imo `next_page` and `previous_page` are perhaps more clear than just `next` and `previous`.

tanema: Thanks, I had the same sort of feeling so it helps to here someone else is thinking the same. I will make that change.

PR #358, opened 2020-02-06
Adding some flexibility into the pagination

### LINKED ISSUE #356: ShopifyResource.find does not propagate kwargs to PaginatedCollection
Hey folks, I'm testing the new pagination and I've found that `resource.find(no_iter_next=True)`
doesn't work as expected (e.g. the resulting iterator will still fetch additional pages).

After having a quick look through the source code, it appears that this is because the `kwargs` are never passed through: https://github.com/Shopify/shopify_python_api/blob/master/shopify/base.py#L213

Overriding the variable on the collection works fine, but is non-ideal:
```python
        collection = resource.find(**kwargs)
        collection._no_iter_next = True
```

### LINKED ISSUE #357: cursor-based pagination not working (header name case-sensitivity?)
I am trying to implement the new cursor-based pagination, with ShopifyPythonAPI/6.0.1 on Python/2.7.16 and pyactiveresource 2.2.0. I am testing with the API version 2020-10.

`has_next_page()` always returns false for me even when there are more items. Looking at your source code, `PaginatedCollection`'s `__parse_pagination` method looks for `metadata['header']['Link']` yet for some reasons, all header names are in lower-cases on my environment.

```
>>> p = shopify.Product.find(limit=3)
>>> p.has_next_page()
False
>>> p.metadata['pagination']
{}
>>> p.metadata['headers']['link']
'<https://STORE.myshopify.com/admin/api/2020-01/products.json?limit=3&page_info=CURSOR>; rel="next"'
>>> p.metadata['headers']['Link']
  File "<stdin>", line 1, in <module>
KeyError: 'Link'
```

on a side note, it would be nice if we could retrieve page_info value from PaginatedCollection, so I can use it to fetch the next page. This way I don't need to keep the original page object to access the next page. I see `shopify.Product.find(page_info=CURSOR)` works, although i would feel better if it was documented. 

Thanks.

tanema:
fixes #357 
fixes #356 

This adds 2 things.

### Link Header
It seems like some versions of urlib mess with the case of header names. I have added in flexibility to this to check for Link and link so that it will work in any case

### no_iter_next
kwargs were not passed to the Paginated collection so there was no way for a user to specify the attribute no_iter_next in the initialization. I have passed along the kwargs now and it can not be specified

@GhostApps @Philpax @dasony for review of changes

### DIFF
--- a/shopify/base.py
+++ b/shopify/base.py
@@ -210,5 +210,5 @@ class ShopifyResource(ActiveResource, mixins.Countable):
         """Checks the resulting collection for pagination metadata."""
         collection = super(ShopifyResource, cls).find(id_=id_, from_=from_, **kwargs)
         if isinstance(collection, Collection) and "headers" in collection.metadata:
-            return PaginatedCollection(collection, metadata={"resource_class": cls})
+            return PaginatedCollection(collection, metadata={"resource_class": cls}, **kwargs)
         return collection
--- a/shopify/collection.py
+++ b/shopify/collection.py
@@ -38,7 +38,7 @@ class PaginatedCollection(Collection):
         self._next = None
         self._previous = None
         self._current_iter = None
-        self._no_iter_next = kwargs.pop("no_iter_next", False)
+        self._no_iter_next = kwargs.pop("no_iter_next", True)
 
     def __parse_pagination(self):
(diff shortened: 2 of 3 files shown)


Philpax:
Looks good to me.

PR #369, opened 2020-03-24
Solves API version +2019-10 fails with variant updates

### LINKED ISSUE #347: API version 2019-10 fails with variant updates
Hello,

I've found what I believe is a bug with the current version of the project (5.1.2).  Targeting API version 2019-10 with Shopify, a call to update the price of a product variant fails with a 400 Bad Request error.  The complete text of the error follows:

pyactiveresource.connection.BadRequest: Response(code=400, body="b'{"error":"Write requests to inventory_quantity and inventory_quantity_adjustment are no longer supported. Please use the Inventory Levels API."}

Steps to reproduce:

1. Create connection via private app key targeting API version 2019-10 (admin/api/2019-10)
2. Set shop resource
3. Set a new object as a product
4. From the product, set another new object as a product variant
5. Call <variant>.price = float(<price>)   (Running print(<variant>.price) does return the updated price here)
6. Call <variant>.save()

Upon calling save(), the above error is thrown.

After reviewing documentation, it appears that inventory_quantity should be removed from the list of variant attributes as it is fully deprecated in API version 2019-10.  This call works without issue targeting API version 2019-07.

JuanjoA:
fixes #347

Hi, I think that PR can solve the problem about updating variants from api versions +2019-10, and the error 400.

### DIFF
--- a/shopify/resources/product.py
+++ b/shopify/resources/product.py
@@ -38,7 +38,9 @@ class Product(ShopifyResource, mixins.Metafields, mixins.Events):
         start_api_version = '201910'
         if api_version >= start_api_version:
             for variant in self.variants:
-                del variant.attributes['inventory_quantity']
-                del variant.attributes['old_inventory_quantity']
+                if variant.attributes.get('inventory_quantity'):
+                    del variant.attributes['inventory_quantity']
+                if variant.attributes.get('old_inventory_quantity'):
+                    del variant.attributes['old_inventory_quantity']
 
         return super(ShopifyResource, self).save()
--- a/shopify/resources/product.py
+++ b/shopify/resources/product.py
@@ -38,9 +38,8 @@ class Product(ShopifyResource, mixins.Metafields, mixins.Events):
         start_api_version = '201910'
         if api_version >= start_api_version:
             for variant in self.variants:
-                if variant.attributes.get('inventory_quantity'):
+                if 'inventory_quantity' in variant.attributes:
                     del variant.attributes['inventory_quantity']
-                if variant.attributes.get('old_inventory_quantity'):
+                if 'old_inventory_quantity' in variant.attributes:
                     del variant.attributes['old_inventory_quantity']
-
         return super(ShopifyResource, self).save()
--- a/shopify/resources/product.py
+++ b/shopify/resources/product.py
@@ -30,3 +30,15 @@ class Product(ShopifyResource, mixins.Metafields, mixins.Events):
     def add_variant(self, variant):
         variant.attributes['product_id'] = self.id
         return variant.save()
+
+    def save(self):
+        # github issue 347
+        # todo: how to get the api version without split & strip
+        api_version = ShopifyResource._site.split('/')[-1].strip('-')
+        start_api_version = '201910'
+        if api_version >= start_api_version:
+            for variant in self.variants:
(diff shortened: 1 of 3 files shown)


shopify/resources/product.py
```diff
@@ -30,3 +30,16 @@ def remove_from_collection(self, collection):
     def add_variant(self, variant):
         variant.attributes['product_id'] = self.id
         return variant.save()
+
+    def save(self):
+        # github issue 347
+        # todo: how to get the api version without split & strip
+        api_version = ShopifyResource._site.split('/')[-1].strip('-')
```
tanema: I believe you should be able to access this directly with `ShopifyResource._version`

JuanjoA: I have tried, but `ShopifyResource._version` has not value.

tanema: What about `self.version`?  This should all be available on ShopifyResources https://github.com/Shopify/shopify_python_api/blob/master/shopify/base.py#L92-L155

(for some reason version is duplicated as well 🤦‍♂ )

JuanjoA: @tanema nop, _ShopifyResource.version_ not return value, and _self.version_ does not exist.

tanema: sorry I should have been more clear. I mean you can actually access it from self. I just tested this myself and it works.

```python
api_version = self.version.strip('-') # this will convert 2020-01 to 202001 which it looks like you are comparing.
```

Also, both the comparison here and the URL stripping do not handle `unstable`

```python
    def save(self):
        start_api_version = '201910'
        if self.version.strip('-') >= start_api_version and self.version != 'unstable':
            for variant in self.variants:
                if 'inventory_quantity' in variant.attributes:
                    del variant.attributes['inventory_quantity']
                if 'old_inventory_quantity' in variant.attributes:
                    del variant.attributes['old_inventory_quantity']
        return super(ShopifyResource, self).save()
```

JuanjoA: When I call on my script:

>         success = shopify_product.save()

And intercept with pdb, I dont have self.version:
```
ShopifyAPI-7.0.1-py3.7.egg/shopify/resources/product.py(40)save()
-> if api_version >= start_api_version:
(Pdb) self.version
*** AttributeError: version

```

tanema: Sorry, still coming to grips with this codebase. It is actually referenced from `_threadlocal` to keep thread saftey.

```
    def save(self):
        api_version = self._threadlocal.version
        start_api_version = '201910'
        if api_version.strip('-') >= start_api_version and api_version != 'unstable':
            for variant in self.variants:
                if 'inventory_quantity' in variant.attributes:
                    del variant.attributes['inventory_quantity']
                if 'old_inventory_quantity' in variant.attributes:
                    del variant.attributes['old_inventory_quantity']
        return super(ShopifyResource, self).save()
```

JuanjoA: Also no value. I think this happens when you dont use session (it's my case). I deleted some repeated lines in _base.py_ and updated getter _get_version_ to retrieve the api version from the path when no value in _version_ or __version_

JuanjoA:
Thank you @tanema for your support.

tanema:
@JuanjoA I just deploy this in version 7.0.2

Issue #386, opened 2020-05-20
Creating a product does not work as documented, need to set variant field


poolebu:
Tried to create a product as documented, gives the following error

![error_001](https://user-images.githubusercontent.com/4500001/82441226-ea51e500-9a9d-11ea-8bed-525302d6c142.png)

Adding variant array works however

![works](https://user-images.githubusercontent.com/4500001/82441261-f9389780-9a9d-11ea-9554-e7346626d6f9.png)

prysie:
![image](https://user-images.githubusercontent.com/12985914/86095001-9ef71180-baf4-11ea-8a73-bafd21e99098.png)

There is a check in the Product save function for API versions >= 201910 it attempts to read the variant attribute, however it has not been set to any value as it is not a required field as per the API specification.

beetree:
I'm having a related issue using version 8.0.1. It can very easily be reproduced, and I've included full code below.

The following code...
```
import shopify
import ssl

ssl._create_default_https_context = ssl._create_unverified_context

keys = {
	"API_KEY": "XXX",
	"PASSWORD": "XXX"
}

shop_url = "https://%s:%s@XXX.myshopify.com/admin/api/2019-10" % (keys["API_KEY"], keys["PASSWORD"])
shopify.ShopifyResource.set_site(shop_url)

shopifyProduct = shopify.Product.find(XXX)
shopifyProduct.save()
```
... runs fine.

Whereas this code (only change is API version)...
```
import shopify
import ssl

ssl._create_default_https_context = ssl._create_unverified_context

keys = {
	"API_KEY": "XXX",
	"PASSWORD": "XXX"
}

shop_url = "https://%s:%s@XXX.myshopify.com/admin/api/2020-01" % (keys["API_KEY"], keys["PASSWORD"])
shopify.ShopifyResource.set_site(shop_url)

shopifyProduct = shopify.Product.find(XXX)
shopifyProduct.save()
```
... throws the following error:
```
Traceback (most recent call last):
  File "/Users/johan.land/PycharmProjects/FarmFoods/bin/tmp.py", line 15, in <module>
    shopifyProduct.save()
  File "/Users/johan.land/PycharmProjects/FarmFoods/venv/lib/python3.8/site-packages/shopify/resources/product.py", line 39, in save
    if 'variants' in self.product.attributes:
[...]

PR #389, opened 2020-06-04
Handle saving of product with no variants specified

rnovacek:
Shopify API allows not setting all the attributes when changing a resource. It can be used as optimization to speed up things if some parts of a resource doesn't change.

Common scenario is to prevent variants update when saving product as there can be up to 100 variants per product and saving all takes a long time.

Recent update to delete `inventory_quantity` and `old_inventory_quantity` of variants when saving product broke the ability to delete variants to prevent them from saving. It was working fine in version 7.0.0.

This PR adds a check if the `variants` are defined on a product before attempt to delete `inventory_quantity` and `old_inventory_quantity` when saving.

### DIFF
--- a/shopify/resources/product.py
+++ b/shopify/resources/product.py
@@ -36,9 +36,10 @@ class Product(ShopifyResource, mixins.Metafields, mixins.Events):
         api_version = ShopifyResource.version
         if api_version and (
                 api_version.strip('-') >= start_api_version) and api_version != 'unstable':
-            for variant in self.variants:
-                if 'inventory_quantity' in variant.attributes:
-                    del variant.attributes['inventory_quantity']
-                if 'old_inventory_quantity' in variant.attributes:
-                    del variant.attributes['old_inventory_quantity']
+            if 'variants' in self.product.attributes:
+                for variant in self.variants:
+                    if 'inventory_quantity' in variant.attributes:
+                        del variant.attributes['inventory_quantity']
+                    if 'old_inventory_quantity' in variant.attributes:
+                        del variant.attributes['old_inventory_quantity']
         return super(ShopifyResource, self).save()


rnovacek:
I've signed the CLA, but I don't know how to refresh it...

tanema:
Thanks!

beetree:
I believe there's an issue with this PR. It throws an error when loading a product, editing it and then saving it. Longer discussion [here](https://github.com/Shopify/shopify_python_api/issues/386).

Reproduce using:

Dockerfile:
```
FROM python:3.8.5
RUN pip install ShopifyAPI
ADD main.py /
```

main.py (example from [here](https://github.com/Shopify/shopify_python_api)):
```
import shopify

keys = {
    "API_KEY": "XXX",
    "PASSWORD": "XXX"
}

shop_url = "https://%s:%s@XXX.myshopify.com/admin/api/2020-07" % (keys["API_KEY"], keys["PASSWORD"])
shopify.ShopifyResource.set_site(shop_url)

product = shopify.Product()
product.title = "Shopify Logo T-Shirt"
print(product.id)
product.save()
```

When running, this throws the error:
```
  File "main.py", line 17, in <module>
    product.save()
  File "/usr/local/lib/python3.8/site-packages/shopify/resources/product.py", line 39, in save
    if 'variants' in self.product.attributes:
  File "/usr/local/lib/python3.8/site-packages/pyactiveresource/activeresource.py", line 917, in __getattr__
    raise AttributeError(name)
AttributeError: product
```

poolebu:
Hello! So I'm back working with the same code, and the issue is still here but in another form. The default stable package at the moment, `8.0.1` does not let me create a product as documented. See bellow

```
API_VERSION = '2020-07'

```

2020-08-01_13h06_05

2020-08-01_13h05_14

If I go back to version `!pip install ShopifyAPI==8.0.0` and `API_VERSION = '2020-04'`  I get the original error, which I can bypass by adding `new_product.variants = []` before saving a new product.

It seems to me that you should test the documented code for creating a new product before releasing a new api version, since its a critical blocker use case, or maybe I'm doing something wrong?

I hope this report is useful, and thank you for the API

beetree:
@rnovacek , any chance you can take a look at this issue? There are multiple reports of it by now.
Thanks!

PR #391, opened 2020-06-09
Add optional customer_id support for Order.find()

metatick:
This PR adds allows users to retrieve orders for a specific customer by specifying a customer_id in the find() call.

The implementation is identical to that of Article


Paulinakhew:
Hey @metatick , I've implemented your suggested change in #466 . Thank you for this PR! 🎉

PR #409, opened 2020-09-22
Update Release notes regarding testing against stable API

andyw8:
We've had some cases (e.g. #386) where the README examples are not in sync with the latest API.

Let's add this step as a reminder to validate them.

(Later we can look into automating this).

### DIFF
--- a/RELEASING
+++ b/RELEASING
@@ -1,18 +1,20 @@
 Releasing shopify_python_api
 
-1. Check the Semantic Versioning page for info on how to version the new release: http://semver.org
+1. Verify that the examples in the README are still valid against the latest stable API release.
 
-2. Update version in shopify/version.py
+2. Check the Semantic Versioning page for info on how to version the new release: http://semver.org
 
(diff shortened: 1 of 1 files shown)


RELEASING
```diff
@@ -1,18 +1,20 @@
...
-5. Tag the release with the version
+6. Tag the release with the version
     git tag -m "Release X.Y.Z" vX.Y.Z
 
-6. Push the changes to github
+7. Push the changes to github
     git push --tags origin master
 
-7. Shipit!
+8. Shipit!
```
andyw8, line 20: (not a markdown file so I had to manually renumber).

mkevinosullivan: Maybe we should just change it to a markdown file, which would align with many of the other repos?

mkevinosullivan (approved):

RELEASING
andyw8: Yeah, makes sense.

PR #416, opened 2020-09-23
Add support for upcoming 2020-10 release, drop 2019-10

andyw8:
This shouldn't be released before Oct 1.


shopify/api_version.py
```diff
@@ -29,6 +29,7 @@ def define_known_versions(cls):
         cls.define_version(Release('2020-01'))
```
tanema, line 29: 2019-10 needs to be removed.

tanema (changes requested):

tanema (approved):

PR #423, opened 2020-10-15
Melaniew/add fulfillment orders coverage

### LINKED ISSUE #420: FulfillmentOrders
# Issue summary

**shopify.FulfillmentOrders.find(fulfillment_order_id, order_id=OrderId)**  I am not able to get result for this method
When calling this method,  I am getting this issue
**Not Found: https://demo.myshopify.com/admin/api/2020-10/orders/2791579517090/fulfillment_orders/2557369221282.json**
## Expected behavior

This should return specific fulfillment order
**https://demo.myshopify.com/admin/api/2020-10/fulfillment_orders/2557369221282.json**

## Actual behavior

It's not returning fulfillment records

- `shopify_python_api` version:
- `pyactiveresource` version:
- Shopify API version used (e.g. `'2020-10'`):

mllemango:
Currently, _FulfillmentOrders_ only hits the endpoint `GET /admin/api/2020-10/orders/{order_id}/fulfillment_orders.json`, but there is a second end point that isn't accessible. 
https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillmentorder

Fixes #420  

This change introduces the option to hit `GET /admin/api/2020-10/fulfillment_orders/{fulfillment_order_id}.json` by checking the arguments passed through. 

- If it contains a `fulfillment_order_id`, use endpoint `/admin/api/2020-10/fulfillment_orders/{fulfillment_order_id}.json`
- If it only contains an `order_id`,  use endpoint `orders/{order_id}/fulfillment_orders`

### Checklist

### DIFF
--- a/shopify/resources/fulfillment.py
+++ b/shopify/resources/fulfillment.py
@@ -15,4 +15,12 @@ class Fulfillment(ShopifyResource):
 
 
 class FulfillmentOrders(ShopifyResource):
-    _prefix_source = "/orders/$order_id/"
+    @classmethod
+    def find(cls, id_=None, from_=None, **kwargs):
+        if id_:
+            cls._prefix_source = ''
+            resource = cls._find_single(id_, **kwargs)
+        else:
+            cls._prefix_source = "/orders/$order_id/"
+            resource = cls._find_every(**kwargs)
+        return resource
--- a/shopify/resources/fulfillment.py
+++ b/shopify/resources/fulfillment.py
@@ -19,8 +19,7 @@ class FulfillmentOrders(ShopifyResource):
     def find(cls, id_=None, from_=None, **kwargs):
         if id_:
             cls._prefix_source = ''
-            resource = cls._find_single(id_, **kwargs)
         else:
             cls._prefix_source = "/orders/$order_id/"
-            resource = cls._find_every(**kwargs)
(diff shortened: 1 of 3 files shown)


andyw8:
I think it might be better to follow the convention already used in shopify_api, where `/admin/api/2020-10/orders/{order_id}/fulfillment_orders.json` is via a method on `Order`

https://github.com/Shopify/shopify_api/blob/6e5eb7f5c8a8ed478abe438976aba4a70f52666b/lib/shopify_api/resources/order.rb#L33

However, this would mean the behaviour of the current `FulfillmentOrder` class would need to change.

Any toughts @tanema ?

tanema:
From the issue
>Not typed URL statically. Its generation after calling this function shopify.FulfillmentOrders.find(fulfillment_order_id, order_id=OrderId) from shopify python library.

It does seem like this developer already has order data, this does indicate to me that this is how fulfillment orders should be accessed through Orders

andyw8:
(I've removed myself from the reviewers for now since we'll need rework).

mllemango:
Closing the PR for now. The suggested fix in #420 as a workaround is acceptable, and rolling out a breaking change for the `FulfillmentOrders` class will require further planning.

PR #428, opened 2020-11-05
[Feature] Add support for Shopify Payments resources

Alchez:

The Shopify API includes endpoints for various [Shopify Payments resources](https://shopify.dev/docs/admin-api/rest/reference/shopify_payments), but they're not currently included and supported in the Python library.

This PR tries to introduce the Balance, Dispute, Payouts and Transactions resources into the library.

### Checklist


paulomarg:
Hi @Alchez, thank you for taking the time to contribute!

At a first glance, the changes you made make sense, but it seems a few tests are failing. Could you please make sure those are passing before we proceed with this PR?

Thank you!

Alchez:
Hey @paulomarg, thanks for reviewing this.

I'm not exactly sure how to resolve the test failures:

- My balance test is failing at trying to retrieve an invalid path that I'm not sure how is being formed.

The path seems to be `GET https://this-is-my-test-show.myshopify.com/admin/api/unstable/shopify_payments/balances.json`, but I haven't mentioned the pluralized version of the resource anywhere.

- The other tests seem to be failing at missing attributes. Is there a place I can configure those?

paulomarg:
For the balance test, by default pyactiveresource will pluralize the item you're looking for when calling `find` as that is the convention for REST interfaces. Since in this case `balance` isn't plural in the API, you'll need to set the `_plural = "balance"` property in your `balance.py` resource.

The other tests seem to be failing because the items found in the `find` calls are not as expected, so you need to make sure the queries are working as expected.

Hope this helps!

Alchez:
@paulomarg I've fixed the balance test, but I'm still unsure about the fixes for the `find` calls.

I've inspected the JSON records for both Payouts and Disputes, and they're in the same format as a few other resources (for example, Article). The only other change I have is the custom `prefix` I've used for the requests, but that seemed to work for the other Payouts and Disputes tests.

What am I doing wrong?

paulomarg:
Hey @Alchez, it seems like you may have an issue with your fixtures, as it's trying to load an array of payloads both when fetching all and when fetching by id - when fetching a single id we expect the response to be a single payload object, rather than an array.

I think you'll need to set up different fixtures for the find and find by id cases.

Hope this helps!

Alchez:
@paulomarg 

Okay, I split the fixtures for single and multiple entries, and the tests are passing now. Thanks for your help!

paulomarg (approved):

mohamedfawzy96:
Hello, Just wanted to ask when will these changes be merged ?

paulomarg:
Hey @mohamedfawzy96, this was actually just merged. These changes will be present in the next release!

mohamedfawzy96:
@paulomarg thank you for the info !

Issue #310, opened 2019-04-09
Updating a Fulfillment via API Wrapper Fails


andessen:
I have been fully utilizing the Python API Wrapper with general success but now I have determined a potential bug when trying to make a PUT request via the API via the fulfillments endpoint.

>>> sff = shopify.Fulfillment.find_first(order_id='932467572816')
>>> ssf.tracking_url = 'http://helloworldtracking.com/some_id'
>>> ssf.save()
>>> ssf.tracking_url
>>>

Here the tracking url is being updated but fails to change via the wrapper. Attempting the same request via a PUT utilizing the endpoint via requests seems to achieve the desired result.

Question: Am I utilizing the API appropriately for fulfillments & PUTs or this is a bug?

https://github.com/Shopify/shopify_python_api

leocxy:
Hi 

I got the same issue. 

Here are my configs

API Version: 2020-07
ShopifyAPI: 8.0.4

Below is my codes

`
fulfills = self.api.Fulfillment.find(order_id=order_id)
            for fulfill in fulfills:
                # check tracking number
                if len(fulfill.tracking_number) == 52 and not fulfill.tracking_url:
                    tracking_url = 'https://www.nzcouriers.co.nz/nzc/servlet/ITNG_TAndTServlet?page=1&VCCA=Enabled&Key_Type=BarCode&barcode_data={}'
                    tracking_url = tracking_url.format(fulfill.tracking_number.replace(' ', '+'))
                    print(tracking_url)
                    fulfill.tracking_url = tracking_url
                    res = fulfill.save()
                    print(res, fulfill.errors.full_messages())
                    exit()
`

The res showing True, and none error messages. But the tracking URL does not update...

mllemango:
Hey @leocxy, in order to update the the tracking_url of the fulfillment, you actually have to use `fulfill.tracking_urls` by assigning a list to it, eg. ` fulfill.tracking_urls = ["google.com"]`. 
Unfortunately this isn't super clear, and I only found this out by looking at our [API documentation](https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillment#update_tracking-2020-01).
Let me know if you're still running into issues updating tracking!

leocxy:
Hi @mllemango, I was using GraphQL instead of "ShopifyAPI" to updating the tracking info. 

I know the API documentation. But I don't know how to call that with the python package "ShopifyAPI". 

Anyway, thank you for your reply.

PR #432, opened 2020-11-11
Add update_tracking endpoint to Fulfillment

mllemango:
The fulfillment API has a clear method of [updating tracking information ](https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillment#update_tracking-2020-01) that our library doesn't support. 
This may help reduce issues such as #310.

Following the footsteps of shopify_api, a new class was created in order to hit the end point `/fulfillments/$fulfillment_id`

### Checklist

### DIFF
--- a/shopify/resources/fulfillment.py
+++ b/shopify/resources/fulfillment.py
@@ -17,7 +17,7 @@ class Fulfillment(ShopifyResource):
     def update_tracking(self, tracking_info, notify_customer):
         fulfill = FulfillmentV2()
         fulfill.id = self.id
-        fulfill.update_tracking(tracking_info, notify_customer)
+        self._load_attributes_from_response(fulfill.update_tracking(tracking_info, notify_customer))
 
 
 class FulfillmentOrders(ShopifyResource):
@@ -34,4 +34,4 @@ class FulfillmentV2(ShopifyResource):
                 "notify_customer": notify_customer
             }
         }
-        self._load_attributes_from_response(self.post("update_tracking", json.dumps(body).encode()))
+        return self.post("update_tracking", json.dumps(body).encode())
--- a/shopify/resources/fulfillment.py
+++ b/shopify/resources/fulfillment.py
@@ -1,4 +1,5 @@
 from ..base import ShopifyResource
+import json
 
 
 class Fulfillment(ShopifyResource):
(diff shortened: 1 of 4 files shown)


shopify/resources/fulfillment.py
```diff
@@ -13,6 +14,24 @@ def complete(self):
...
+    def update_tracking(self, tracking_info, notify_customer):
+        fulfill = FulfillmentV2()
+        fulfill.id = self.id
+        self._load_attributes_from_response(fulfill.update_tracking(tracking_info, notify_customer))
+
 
 class FulfillmentOrders(ShopifyResource):
     _prefix_source = "/orders/$order_id/"
+
+class FulfillmentV2(ShopifyResource):
```
paulomarg, line 26: I'm not sure I follow the idea of the `V2` nomenclature, is there a particular reason we can't just do this in `Fulfillment`? If the class is absolutely necessary, we might want to consider renaming it to something that better conveys what it does.

mllemango: I don't _believe_ we can do it in the original `Fulfillment` class because that class points to `/orders/$order_id/fulfillment/$fulfillment_id`, and the endpoint in this case is only `fulfillment/$fulfillment_id`. I would also prefer to do it in 1 class, but I'm not quite sure how..
The `V2` nomenclature is from [shopify_api](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/fulfillment_v2.rb) (that is used by [Fulfillment.rb](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/fulfillment.rb)) in a similar manner. Thought I would keep it consistent between the two libs

paulomarg: All of that makes perfect sense, thanks for the context! I'm still not a fan of calling it `V2` but I agree we should keep them consistent.

Not saying we should do this, but I wonder if it wouldn't be cleaner to change `_prefix_source` for that particular method rather than adding a whole new class for it.

paulomarg (approved):

shopify/resources/fulfillment.py
mllemango: I tried manipulating `_prefix_source` but wasn't successful with my attempt. I do believe the convention of changing prefixes is to override the classmethod `_prefix` depending on passed in options, but in this case the usage is after we define a Fulfillment object already :/
Thus new class

paulomarg: Nothing wrong with this approach, I was just thinking aloud there!

paulomarg (approved):
LGTM

PR #434, opened 2020-11-18
Add github actions 

mllemango:
This library is running on travis-ci.org, [which will be going into read-only mode at the end of the year](https://docs.travis-ci.com/user/migrate/open-source-repository-migration/#frequently-asked-questions)
> we are announcing that travis-ci.org will be officially closed down completely no later than December 31st, 2020

Taking this opportunity to move away from travis-ci all together, and onto Github Actions. 
It is my understanding that we aren't billed for action usage on public repos.

Covering the CI pipeline with Github Actions. 

### Checklist


.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Python ${{ matrix.version }}
+    strategy: 
+      matrix:
+        version: [2.7, 3.4, 3.5, 3.6]
```
mllemango, line 13: Wondering if we should pop in more versions to test?

andyw8: We'll keep 2.7, and for the 3 series I think we should cover the latest release, and the oldest supported, so 3.6 and 3.9. I don't think we need to cover the in between ones.

andyw8:
Let's also remove the `.travis.yml`

.github/workflows/build.yml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.version }}
+      - name: Install Dependencies
+        run: | 
+          python -m pip install --upgrade pip
+          pip install flake8 pytest
+          pip install mock
```
andyw8, line 25: Any idea why the travis config didn't need these additional steps?

mllemango: Really not sure! :/

andyw8 (approved):

PR #441, opened 2020-12-02
[Feature] Add Dynamic api versioning

### LINKED ISSUE #339: Dynamic API Version support
Similar to https://github.com/Shopify/shopify_api/pull/600, so we don't have to manually update the gem on every API Version release.

Worth noting that the description of that PR is somewhat different from the final approach chosen.

smubbs:
This PR will automatically populate with the current API versions supported for the Shopify Admin API.

Hopefully, this should remove the requirement to manually update the list of current API versions.

Fixes #339 

Versioning has been here a while and it's probably time we fix this manual update requirement. A manual system is fine, but it still causes issues with development and to not be able to select the currently available APIs, that is to say, to have to wait for this package to be updated every quarter is unreasonable. It is now 25 November, Release Candidate 2021-01 has been available for also 2 months now and it is still not supported here.

`ApiVersion.define_known_versions()` has been changed to make a request to https://app.shopify.com/services/apis.json and automatically populate the versions list.

Updated `TestCase.fake()` default to using an `url` when provided or constructing one when `endpoint` is provided.

Added payload `api_version.json` for call in `fake('apis'...)`
### Checklist


mllemango (approved):
Thank you for working on this! I've tested locally and this looks good!

smubbs:
Thanks! Happy to help.

tanema (approved):

andyw8:
(note that we're missing a change entry for this, even though the checkbox was checked).

mllemango:
Good catch, I'll add an entry with the new release PR if thats fine.

PR #451, opened 2021-01-05
Release v8.2.0

mllemango:

Releasing 8.2.0 with changelog update

### Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+== Version 8.2.0
+- [Feature] Add support for Dynamic API Versioning (#441)
+
 == Version 8.1.0
 - [Feature] Add support for Shopify Payments resources (#428)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG
```diff
@@ -1,3 +1,6 @@
+== Version 8.2.0
+- [Feature] Add support for Dynamic API Versioning (#441)
```
paulomarg, line 2: Can we make the PR a link? Makes it easier to drill into changes later.

paulomarg (approved):
Looks good other than the link thing.

CHANGELOG
```diff
@@ -1,3 +1,6 @@
+== Version 8.2.0
+- [Feature] Add support for Dynamic API Versioning ([#441](https://github.com/Shopify/shopify_python_api/pull/441))
```
andyw8, line 2: I think it would be worth expanding this, e.g.:

> Add support for Dynamic API Versioning. When the library is initialized, it will now make a request to Shopify to fetch a list of the available API versions.

(this may be useful information to know if someone is using [VCR.py](https://github.com/kevin1024/vcrpy) or similar in their test suite).

mllemango: I see! That makes sense!

andyw8 (approved):

Issue #361, opened 2020-02-20
How to create a FulfillmentEvent


lswamy:
I can create and update fulfillments through the api to add a tracking number to the order when it ships. However the order status does not update for the customer, it just says order is confirmed and come back for updates.
The docs indicate that creating a FulfillmentEvent will update the order status page.

https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillmentevent

How can I accomplish this with the python api?

Thanks

tanema:
Currently there is no fulfillment event resource in the library. Nor does the Fulfillments model have the events mixin. You could, however, extend the Event model or create your own model in your codebase to support this until support is added. Just inherit from ShopifyResource

lswamy:
Thanks for the response.
I created my own model, but I'm not too familiar with pyactiveresource so it may not be correct. It appears to work for my purpose. If this looks good you can close the ticket.

```python
import shopify

class FulfillmentEvent(shopify.ShopifyResource):
    _prefix_source = "/orders/$order_id/fulfillments/$fulfillment_id"
    _singular = "event"
    _plural = "events"

    @classmethod
    def _prefix(cls, options={}):
        order_id = options.get("order_id")
        fulfillment_id = options.get('fulfillment_id')
        if order_id:
            return "%s/orders/%s/fulfillments/%s" % (
                cls.site, order_id, fulfillment_id)
        else:
            return cls.site

event = FulfillmentEvent({'order_id': order_id, 'fulfillment_id': fulfillment_id})
event.status = "in_transit"
event.save()
```

rlepore:
@tanema It doesn't look like this is added yet so I've been running some tests using a custom `FulfillmentEvent` resources like above.  I noticed an issue when the resources returns an error.  For example, if you send an invalid `status` the client throws an unhandled Exception.  It looks like the pyactiveresource library can't parse the response properly.  

Here is the stacktrace along with the error:
```
  File "/lib/python3.7/site-packages/pyactiveresource/activeresource.py", line 453, in create
    resource.save()
  File "/lib/python3.7/site-packages/pyactiveresource/activeresource.py", line 850, in save
    self.errors.from_json(err.response.body)
  File "/lib/python3.7/site-packages/pyactiveresource/activeresource.py", line 132, in from_json
    self.from_array(errors)
  File "/lib/python3.7/site-packages/pyactiveresource/activeresource.py", line 82, in from_array
    attr_name = message.split()[0]
[...]

PR #454, opened 2021-01-29
Melaniewpaulinak/add fulfillment event

mllemango:

Fixes https://github.com/Shopify/shopify_python_api/issues/361

Introducing new resource FulfillmentEvent that maps to the fulfillment_event api https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillmentevent

### Checklist

### DIFF
--- a/shopify/resources/fulfillment_event.py
+++ b/shopify/resources/fulfillment_event.py
@@ -1,19 +1,18 @@
 from ..base import ShopifyResource
 
 class FulfillmentEvent(ShopifyResource):
-    _prefix_source = "/orders/$order_id/fulfillments/$fulfillment_id"
-    _singular = "event"
-    _plural = "events"
+    _prefix_source = "/orders/$order_id/fulfillments/$fulfillment_id/"
+    _singular = 'event'
+    _plural = 'events'
 
     @classmethod
     def _prefix(cls, options={}):
         order_id = options.get("order_id")
         fulfillment_id = options.get('fulfillment_id')
-        if order_id:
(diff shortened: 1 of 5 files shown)


shopify/resources/fulfillment_event.py
```diff
@@ -0,0 +1,21 @@
...
+        order_id = options.get("order_id")
+        fulfillment_id = options.get('fulfillment_id')
+        event_id = options.get("event_id")
+
+        return "%s/orders/%s/fulfillments/%s" % (
+            cls.site, order_id, fulfillment_id)
+
+    def save(self):
+      status = self.attributes['status']
+      if status not in ['label_printed', 'label_purchased', 'attempted_delivery', 'ready_for_pickup', 'picked_up', 'confirmed', 'in_transit', 'out_for_delivery', 'delivered', 'failure']:
```
mllemango, line 19: Unsure if hardcoding a list of accepted status' is the best way to go about this

thecodepixi: As for whether to hard code them: Are these likely to change in the future at all/frequently? And is there another way to reference them?

mllemango: No other way to reference status as far as I know

thecodepixi: Yeah I didn't think so, but wanted to ask. I would maybe just assign a variable for the array but otherwise if that's the only option, that's the only option 🤷

tanema (approved):
Looks good to me

PR #456, opened 2021-02-08
Fix 'object of type "filter" has no len()' issue in shopify_api.py

### LINKED ISSUE #426: python 3.8  shopify_api.py line 55, TypeError: object of type 'filter' has no len()
# Issue summary

Write a short description of the issue here ↓

## Expected behavior

What do you think should happen?

## Actual behavior

What actually happens?

Tip: include an error message (in a `<details></details>` tag) if your issue is related to an error

## Steps to reproduce the problem

1.
1.
1.

## Reduced test case

The best way to get your bug fixed is to provide a [reduced test case](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Reducing_testcases).

## Specifications

- `shopify_python_api` version:
- `pyactiveresource` version:
- Shopify API version used (e.g. `'2020-07'`):

Paulinakhew:

Fixes #426 

- previously, the code tried to get the length and index of a `filter`
- if you print out a filter, it will look like this
```
>>> f
<filter object at 0x10b0cedc0>
```

- this filter object has no `len()` attribute
```
>>> len(f)
  File "<stdin>", line 1, in <module>
TypeError: object of type 'filter' has no len()
```

- by converting the filter into a list, we will be able to get the length and index of a filter

### Checklist

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 - Added support for Fulfillment.update_tracking ([#432](https://github.com/Shopify/shopify_python_api/pull/432))
 - Add FulfillmentEvent resource ([#454](https://github.com/Shopify/shopify_python_api/pull/454))
+- Fix for being unable to get the len() of a filter ([#456](https://github.com/Shopify/shopify_python_api/pull/456))
 
 == Version 8.2.0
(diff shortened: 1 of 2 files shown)


mllemango (approved):
LGTM!

PR #457, opened 2021-02-08
Add application credit

### LINKED ISSUE #343: Missing support for ApplicationCredit
I need to create ApplicationCredit objects, but there is no such resource in this SDK

Paulinakhew:

Fixes #343

- introduce new resource ApplicationCredit that maps the the [application_credit api](https://shopify.dev/docs/admin-api/rest/reference/billing/applicationcredit)

### Checklist

### DIFF
--- a/shopify/resources/application_credit.py
+++ b/shopify/resources/application_credit.py
@@ -1,12 +1,4 @@
 from ..base import ShopifyResource
 
 class ApplicationCredit(ShopifyResource):
-    _prefix_source = "application_credits/$application_credit_id/"
-
-    @classmethod
-    def _prefix(cls, options={}):
-        application_credit_id = options.get("application_credit_id")
-        if application_credit_id:
-            return "%s/application_credits/%s" % (cls.site, application_credit_id)
-        else:
-            return "%s" % (cls.site)
+    pass
--- a/shopify/resources/application_credit.py
+++ b/shopify/resources/application_credit.py
@@ -1,0 +1,12 @@
+from ..base import ShopifyResource
+
+class ApplicationCredit(ShopifyResource):
+    _prefix_source = "application_credits/$application_credit_id/"
+
+    @classmethod
+    def _prefix(cls, options={}):
+        application_credit_id = options.get("application_credit_id")
+        if application_credit_id:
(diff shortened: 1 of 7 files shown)


shopify/resources/application_credit.py
```diff
@@ -0,0 +1,12 @@
...
+class ApplicationCredit(ShopifyResource):
+    _prefix_source = "application_credits/$application_credit_id/"
+
+    @classmethod
+    def _prefix(cls, options={}):
+        application_credit_id = options.get("application_credit_id")
+        if application_credit_id:
+            return "%s/application_credits/%s" % (cls.site, application_credit_id)
+        else:
+            return "%s" % (cls.site)
```
Paulinakhew, line 12: tbh I'm not sure if this part is right... this was how I got it to return the site with `/application_credits` tacked onto the end

shopify/resources/application_credit.py
```diff
@@ -0,0 +1,12 @@
+from ..base import ShopifyResource
+
+class ApplicationCredit(ShopifyResource):
+    _prefix_source = "application_credits/$application_credit_id/"
```
mllemango, line 4: I don't believe you need to specify this and add logic for the _prefix, pyactiveresource will default to `application_credits/$id` naturally. A `pass` here should still work

shopify/resources/application_credit.py
jamiemtdwyer: Overloading the `_prefix` isn't necessary here, it's only needed for resources that can be conditionally accessed at different routes, e.g. variants can be accessed from `/admin/#{product_id}/#{variant_id}`, or simply `/admin/#{variant_id}`.

shopify/resources/application_credit.py
jamiemtdwyer: +1 for `pass`, check out the implementation in the Ruby client library: https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/resources/application_credit.rb

`pyactiveresource` should be able to do all of the heavy lifting here with no special logic

Paulinakhew: Okay thank you so much @mllemango @jamiemtdwyer

mllemango (approved):
Love the tests!

PR #459, opened 2021-02-09
Add coveralls

Paulinakhew:

Implements coveralls, which fails a build if coverage has decreased
- this is useful is someone plans to push code they have not tested

- add coveralls

### Checklist

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,2 @@
 [run]
-include = shopify/resources/
+omit = test/*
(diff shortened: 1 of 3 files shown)


andyw8:
We use a similar tool, https://www.codecov.io/, for Theme Kit – any particular reason for choosing Coveralls here?

Paulinakhew:
> We use a similar tool, https://www.codecov.io/, for Theme Kit – any particular reason for choosing Coveralls here?

I've used coveralls before so I decided to implement it! I think sometime today I'll also implement codecov so we can compare them 😄

Paulinakhew:
Closed because #460 was merged in 🥳

PR #460, opened 2021-02-10
Add codecov to measure code coverage across repo

Paulinakhew:
- need a way to see code coverage

- implement codecov

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,12 +1,5 @@
 name: CI
-
-env:
-  CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-
-on: 
-  push:
-
-
+on: [push]
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -21,8 +24,10 @@ jobs:
...
           python -m pip install --upgrade pip
-          pip install pytest mock
+          pip install pytest mock codecov
           python setup.py install 
       - name: Run Tests
-        run: python -m pytest -v
-
+        run: |
+          coverage run -m pytest -v
+      - name: Upload to codecov.io
```
andyw8, line 32: Perhaps we can use https://github.com/codecov/codecov-action for this?

Paulinakhew:
@andyw8 do you have any idea why it's failing? 😢

andyw8:
Is it perhaps because we need to run `python setup.py install` before `pytest --cov=./ --cov-report=xml`?

Paulinakhew:
> Is it perhaps because we need to run `python setup.py install` before `pytest --cov=./ --cov-report=xml`?

😀 yes thank you that fixed it

andyw8 (approved):

PR #462, opened 2021-02-11
Add pre-commit

Paulinakhew:
- add [pre-commit](https://pre-commit.com/) which checks your files before you commit them
- if checks fail, you will not be able to push the code
image

image

- all of the code changes were a result of
```
pre-commit run --all-files
``` 
- no code changes were made, the changes were simply whitespace changes or removing unused imports

### Checklist


.pre-commit-config.yaml
```diff
@@ -0,0 +1,19 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.2.0
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-yaml
+    -   id: check-added-large-files
```
andyw8, line 10: Is there a way to enforce these checks as part of CI?

bin/shopify_api.py
```diff
@@ -8,6 +8,6 @@
 project_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.insert(0, project_root)
 
-with open(os.path.join(project_root, 'scripts', 'shopify_api.py')) as f:
-    code = compile(f.read(), f.name, 'exec')
+with open(os.path.join(project_root, "scripts", "shopify_api.py")) as f:
```
andyw8, line 11: Which check causes the quotes to change?

Paulinakhew: that is done by [black](https://github.com/psf/black) which calls itself "the uncompromising code formatter". If we don't like this I can add `--skip-string-normalization` as an argument

PR #465, opened 2021-02-22
Add autopep8

Paulinakhew:

Fixes #0000 

- autopep8 automatically formats Python code to conform to the PEP 8 style guide

- add autopep8, which is used in [shopify python standards](https://github.com/Shopify/shopify_python)
  - added max line length of 120
- the changes are non-invasive and are mostly whitespace changes

### Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,3 +6,7 @@ repos:
     hooks:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
+-   repo: https://github.com/pre-commit/mirrors-autopep8
+    rev: v1.5.4
+    hooks:
+    -   id: autopep8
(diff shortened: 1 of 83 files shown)


paulomarg:
A few questions before I 🎩  this one:
- Were all of the changes aside from the pre-commit hook made automatically?
- Since we're using a remote repo for the hook, I assume this doesn't introduce any dependencies?

Paulinakhew:
> Were all of the changes aside from the pre-commit hook made automatically?

@paulomarg yup everything but the `.pre-commit-config.yaml` and `pyproject.toml` files were made automatically. `pyproject.toml` is just a config file that `autopep8` uses to auto-format all of the files

> Since we're using a remote repo for the hook, I assume this doesn't introduce any dependencies?

Yup! I just ran pre-commit locally to make the changes automatically. Going forward developers don't need pre-commit locally (unless they want it) because the pre-commit GitHub action will tell them about linting errors and they can fix it themselves

paulomarg (approved):
Tried this locally, seems to work just fine!

PR #467, opened 2021-02-24
Add black hook and run on all files

Paulinakhew:
- add [black](https://github.com/psf/black), _the uncompromising Python code formatter_
  - this makes this project fully adhere to the [Shopify Python](https://github.com/Shopify/shopify_python) guide (_with the exception of quotes_, see below)

> By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.
> Blackened code looks the same regardless of the project you're reading. Formatting becomes transparent after a while and you can focus on the content instead.
> Black makes code review faster by producing the smallest diffs possible.

#### Quotes
- important note: I added `skip-string-normalization` in the `pyproject.toml` file, as changing all the quotes to double quotes is:
  - introducing lots of changes in a single PR
  - slightly unnecessary
  - something that needs discussion with the team
- if you want to read more about why black uses double quotes, you can click [here](https://black.readthedocs.io/en/stable/the_black_code_style.html#strings)
  - this makes a lot of sense to me so if the team agrees I can turn off `skip-string-normalization` in a future PR
  - Shopify/Google style guidelines wants users to be consistent with choice of string quote character (so either single or double) and right now we are using both 💀❌

### Checklist


mllemango:
> It requires Python 3.6.0+ to run but you can reformat Python 2 code with it, too.

Does this mean users getting started with this repo are locked into 3.6+?

paulomarg:
Couple of questions:
- Does the Shopify (or Google) standard specify black as the tool to use for formatting the code? If not, any particular reason to use it?
- Adding to @mllemango's question, will the pre-commit hooks change the code or raise errors if there are infractions?

Paulinakhew:
> Does this mean users getting started with this repo are locked into 3.6+?
@mllemango no it's in CI so users don't have to worry about python 3.6 or installing it (it's not a local dependency). Black can also be used to reformat python 2 code so if we have any outdated code it won't break

Paulinakhew:
> Does the Shopify (or Google) standard specify black as the tool to use for formatting the code? If not, any particular reason to use it?

The Google style guide explicitly says that we should use `pylint`, which is a static code checker. Before we add pylint, I've already added autopep8, and I'm currently adding black, as these two tools automatically format code. Since `pylint` is static, it only offers suggestions and warnings (which can pile up and will be impossible to do manually).

> Will the pre-commit hooks change the code or raise errors if there are infractions?

If I'm running pre-commit locally, it can change the files and fix them before I do `git push`. It normally only runs on files that have been changed in a single commit. I ran `pre-commit run --all-files` on each pre-commit branch so that we'll have a clean code base. Developers that don't have pre-commit installed locally won't have to deal with reformatting everything manually. To answer your question, the hooks can do both and it depends on where `pre-commit` is being run. The hooks that I've enabled so far (`end-of-file-fixer`, `black`, etc.) are ones that _change_ the code if I'm using pre-commit locally. Since pre-commit is in CI, it is unable to change the code and it only raises errors if there are infractions.

paulomarg (approved):
I think this works as a step towards getting us to comply with pylint, but I'd put replacing autopep + black with pylint in the future, and then run pylint as part of CI.

mllemango (approved):

PR #471, opened 2021-03-01
Run black to format all quotes

Paulinakhew:

- a lot of files have both single quotes `'` and double quotes `"`
- this results in linting errors, for example `pylint` results in the following
image
- the [Shopify Python](https://github.com/Shopify/shopify_python) guide uses the [Google Python style guide](https://google.github.io/styleguide/pyguide.html) which asks users to [run pylint](https://google.github.io/styleguide/pyguide.html)
  - since we need to run Pylint, we need to fix the `inconsistent-quotes` errors first

- the only change I've made is to remove `skip-string-normalization = true` from `pyproject.toml`
- `black` automatically formats all of the quotes to double quotes, unless it would introduce more backslash escapes
- e.g. in the following example, it replaces the backslash escapes with outer single quotes
image

- an empty string in single quotes `''` can be confused with a single double quote symbol `"`
- double quotes are consistent with C, which Python interacts a lot with
- for more info read the [Black documentation on strings](https://black.readthedocs.io/en/stable/the_black_code_style.html#strings)

### Checklist

### DIFF
--- a/bin/shopify_api.py
+++ b/bin/shopify_api.py
@@ -8,6 +8,6 @@ import os.path
 project_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.insert(0, project_root)
 
-with open(os.path.join(project_root, 'scripts', 'shopify_api.py')) as f:
-    code = compile(f.read(), f.name, 'exec')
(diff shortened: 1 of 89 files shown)


mllemango (approved):
LGTM

PR #473, opened 2021-03-01
Add pylint

Paulinakhew:
- add [pylint](https://docs.pylint.org/en/1.6.0/intro.html), a tool for checking code 

- adding pylint as a pre-commit hook that is used in CI
- `pylint` is used by the [Google Python Style guide](https://google.github.io/styleguide/pyguide.html) which is used in the [Shopify Python guide](https://github.com/Shopify/shopify_python)
- the `pylintrc` file added is the one from the Google style guide with the exception of:
  - `max-line-length=120` (Shopify style guide asks for 120)
  - `confidence=INFERENCE` (ignore small changes that pylint isn't too confident about)
  - `ignore=missing-class-docstring,missing-module-docstring` (most modules/classes don't have docstrings)
- I manually added a couple of changes


.pre-commit-config.yaml
```diff
@@ -14,3 +14,7 @@ repos:
     rev: 20.8b1
     hooks:
     -   id: black
```
paulomarg, line 16: Do we still need black / autopep? Not a problem if we do, just checking!

paulomarg (approved):
This all makes sense to me!

.pre-commit-config.yaml
Paulinakhew: I think we still do so future PRs have the same formatting that black/autopep suggest! Black has a very specific code style so enforcing that going forward makes the code more readable

paulomarg: That makes sense.

PR #479, opened 2021-03-10
[Feature] Add session token support

NabeelAhsen:

With session tokens now a requirement for Shopify apps on the Shopify App Store, we're trying to support session tokens across all of our third-party and open-source libraries.

This PR introduces the following changes:

1. Create a `SessionTokenUtility` object with a `get_decoded_session_token` public method
    1. Developers can interface with this by passing in a raw `HTTP_AUTHORIZATION_HEADER` along with their app's `api_key` and `secret`
    1. This method validates the existence, expiry, and payload claims of the session token found in the provided header
1. Creates a library dependency on [PyJWT](https://pypi.org/project/PyJWT/)
    1. PyJWT `v2.0.0` is not compatible with python 2.7, and so environment markers are used in `setup.py` to conditionally install the supported versions.
1. Removes `autopep8` from the pre-commit configuration
    1. `autopep8` and `black` format the files in a contradicting manner
    1. The author of `pre-commit` advises against having multiple formatters here: https://stackoverflow.com/a/61687776

### Recommended usage

Developers can use this utility to create decorators for their Python apps. Here's an example usage for a Django API:

```py
from shopify import session_token


def with_session_token(func):
    def wrapper(*args, **kwargs):
        request = args[0]
        try:
            session_token = session_token.decode_from_header(
                authorization_header = request.META.get("HTTP_AUTHORIZATION"),
                api_key = SHOPIFY_API_KEY,
                secret = SHOPIFY_API_SECRET
            )

            # ... do something with the session token payload

            return func(*args, **kwargs)
        except Exception:
            return HttpResponse(status=401)

@with_session_token
def products(request):
    # ... protected resource
```

### Checklist


.pre-commit-config.yaml
```diff
@@ -6,10 +6,6 @@ repos:
     hooks:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
--   repo: https://github.com/pre-commit/mirrors-autopep8
-    rev: v1.5.4
-    hooks:
-    -   id: autopep8
```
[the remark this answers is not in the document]
NabeelAhsen: Here's a sample failing build: https://github.com/Shopify/shopify_python_api/runs/2080197934?check_suite_focus=true

test/utils_test.py
```diff
@@ -0,0 +1,92 @@
...
+            "aud": self.api_key,
+            "sub": "1",
+            "exp": timestamp((current_time + timedelta(0, 60))),
+            "nbf": timestamp(current_time),
+            "iat": timestamp(current_time),
+            "jti": "4321",
+            "sid": "abc123",
+        }
+
+    def test_SessionTokenUtility_getDecodedSessionToken_raises_if_auth_header_is_not_Bearer(self):
```
Paulinakhew, line 42: can you move all of the functions with the `test_SessionTokenUtility_getDecodedSessionToken` prefix into its own class? So it would be something like
```python3
class TestSessionTokenUtilityGetDecodedSessionToken(TestCase):
    def test_raises_if_auth_header_is_not_Bearer(self):
        pass
```

or you can just change
```python3
class UtilsTest(TestCase):
```
on line 21 to the new class name

.pre-commit-config.yaml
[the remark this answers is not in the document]
Paulinakhew: Yeah this makes sense, I prefer `black` over `autope8` as well since it's more detailed

test/utils_test.py
```diff
@@ -0,0 +1,92 @@
...
+    import time
+
+    def timestamp(date):
+        return time.mktime(date.timetuple())
+
+
+else:
+
+    def timestamp(date):
+        return date.timestamp()
```
Paulinakhew, line 18: Can we do
```suggestion
if sys.version_info[0] < 3:  # Backwards compatibility for python < v3.0.0
    import time

def timestamp(date):
    return time.mktime(date.timetuple()) if sys.version_info[0] < 3 else date.timestamp()
```
instead?

test/utils_test.py
NabeelAhsen: For sure! I'd also love to hear your thoughts on how we can organize this file + class for extensibility. Ideally, I foresee adding more utility methods + classes in the future. 

I'm unsure whether they should belong in this file, or in other files under `/utilities` or something. Do you have any thoughts on this?

Paulinakhew: Now that you mention it, our tests are not sorted at all 😆 ideally, we would have our tests mirror the directory structure of the other files, especially since we're introducing `utils_test.py` in `shopify/`. It would look like this
`shopify/resources/`=> `test/shopify/resources/`
`shopify/utils_test.py` => `test/shopify/`
I would want a second opinion from @mllemango but I think you can leave the test in `test/` and later we can open a PR to move all the tests.

In terms of organizing the classes, I like creating a new class per function that we test. So the classes would be like
```python3
class TestNameOfFunction(TestCase):
    def test_success(self):
        pass
```

rezaansyed (approved):
LGTM

mllemango (approved):
LGTM! Love the tests

Paulinakhew (approved):

NabeelAhsen:
Latest commit restructures this into a subpackage:

```
> shopify_app
  * __init__.py
  > resources
  > utils
    * __init__.py
    * utilities.py
    * exceptions.py
> test
  > utils
    * utilities_test.py
```

paulomarg (approved):
🚀

shopify/utils/utilities.py
```diff
@@ -0,0 +1,64 @@
...
+import re
+import sys
+
+if sys.version_info[0] < 3:  # Backwards compatibility for python < v3.0.0
+    from urlparse import urljoin
+else:
+    from urllib.parse import urljoin
+
+
+class SessionTokenUtility:
```
greatestape, line 13: One difference between Python and Ruby is that you can declare functions in Python. I'd be interested in hearing from other Python folks, but my feeling is that if the class doesn't have any instance methods, it doesn't need to be a class. The contents of this class could become module-level constants and public and private functions. I don't think Python has any concept of "private" functions in terms of blocking access, but the naming convention makes the intention clear, which is sufficient IMO

rezaansyed: So we can have something like a `session_token_utility.py` module that contains the necessary functions?

shopify/utils/utilities.py
```diff
@@ -0,0 +1,64 @@
...
+
+    @classmethod
+    def __validate_issuer(cls, decoded_payload):
+        cls.__validate_issuer_hostname(decoded_payload)
+        cls.__validate_issuer_and_dest_match(decoded_payload)
+
+    @classmethod
+    def __validate_issuer_hostname(cls, decoded_payload):
+        hostname_pattern = r"[a-z0-9][a-z0-9-]*[a-z0-9]"
+        shop_domain_re = re.compile(r"^https://{h}\.myshopify\.com/$".format(h=hostname_pattern))
```
greatestape, line 51: I think it'd be worth building this re object once and keeping it in a constant

shopify/utils/exceptions.py
```diff
@@ -0,0 +1,10 @@
+class InvalidIssuerError(Exception):
+    pass
+
+
+class MismatchedHostsError(Exception):
+    pass
+
+
+class TokenAuthenticationError(Exception):
+    pass
```
greatestape, line 10: I wonder if we should either just use one exception class, or have a parent class that these inherit from. The value of that is that it enables adopters to catch all session token validation related errors with one catch block.

To fully realize the benefits of this, we'd probably also want to catch whatever errors PyJWT raises and re-raise so that they're under the same exception hierarchy. There's a handy `raise <new exception> from <some other exception>` format that ensures the original stack trace is preserved.

shopify/utils/utilities.py
```diff
@@ -0,0 +1,64 @@
+from .exceptions import *
```
greatestape, line 1: I think * imports are frowned upon. I tend to like the pattern you've got below where you import the module and then access its members:

```python
from . import exceptions

...

    raise exceptions.InvalidIssuerError

...
```

shopify/utils/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .exceptions import *
+from .utilities import SessionTokenUtility
```
greatestape, line 2: I'm -1 on "utils", "utilities", "tools", "extra", "misc", "helpers", etc.

It's not a hill I always want to die on, but I think a name that helps clarify what should go in and what shouldn't is always nicer. In this case, maybe `session_tokens`?

Also, the exceptions aren't too cumbersome. It would simplify things if we just kept everything in `session_tokens.py`.

test/utils/utilities_test.py
```diff
@@ -0,0 +1,87 @@
...
+            SessionTokenUtility.get_decoded_session_token(
+                self.build_auth_header(), api_key=self.api_key, secret=self.secret
+            )
+
+    def test_returns_decoded_payload(self):
+        decoded_payload = SessionTokenUtility.get_decoded_session_token(
+            self.build_auth_header(), api_key=self.api_key, secret=self.secret
+        )
+
+        self.assertEqual(self.payload, decoded_payload)
```
greatestape, line 87: Could we also test that the session token is rejected if it's not properly signed or if it's not using the right encoding (for example, unsigned)?

shopify/utils/exceptions.py
paulomarg: I like having the shared base class for exceptions as it helps callers know whether the exception is coming from us or not.

paulomarg (approved):
I like the module changes, it looks a lot cleaner like this!

README.md
```diff
@@ -120,6 +120,60 @@ _Note: Your application must be public to test the billing process. To test on a
...
+decoded_payload = session_token.get_decoded_session_token(
+    authorization_header=your_auth_request_header,
+    api_key=your_api_key,
+    secret=your_api_secret,
+)
+```
+
+#### Use session_token to create a decorator
+
+Here's a sample use-case of building a Django decorator using `session_token`:
```
rezaansyed, line 141: If we're providing a Django example here, we should also be providing a Flask example. The library is supposed to be framework agnostic and the usage should cover popular frameworks.

rezaansyed (approved):
Minor comment but looks good otherwise!

NabeelAhsen:
Tested locally on a sample Django app, and everything works! Feeling confident to ship.

PR #482, opened 2021-03-11
Introduce ApiAccess for representing access scopes

rezaansyed:

There are common operations that are used for managing access scopes in apps. Such operations include serializing, deserializing and normalizing scopes. 

This PR introduces the `ApiAccess` value object that will be used to perform such operations. This helps ensure that apps do not potentially write duplicate code for scopes and delegating this responsibility to the `shopify_python_api`.

For further reference, we made a similar move in the [Shopify API ruby gem](https://github.com/Shopify/shopify_api/blob/master/lib/shopify_api/api_access.rb)

#### Constructing `ApiAccess`
```python
api_access = ApiAccess(["read_products", "write_orders"])
another_api_access = ApiAccess("read_products, write_products")
```

#### Delegating check for whether scopes are mismatched or subset of another
```python
def scopes_match?(existing_scopes, requesting_scopes)
  expected_api_access = ApiAccess(existing_scopes)
  actual_api_access = ApiAccess(requesting_scopes)

  actual_api_access == expected_api_access
```

```python
def scopes_subset?(existing_scopes, requesting_scopes)
  expected_api_access = ApiAccess(existing_scopes)
  actual_api_access = ApiAccess(requesting_scopes)

  actual_api_access.covers(expected_api_access)
```

####

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,24 @@ To easily install or upgrade to the latest release, use [pip](http://www.pip-ins
 pip install --upgrade ShopifyAPI
 ```
 
+### Table of Contents
+
+- [Getting Started](#getting-started)
+  - [Public and Custom Apps](#public-and-custom-apps)
+  - [Private Apps](#private-apps)
+- [Billing](#billing)
+- [Session Tokens](docs/session-tokens)
+- [Handling Access Scope Operations](docs/api-access.md)
+- [Advanced Usage](#advanced-usage)
+- [Prefix Options](#prefix-options)
+- [Console](#console)
+- [GraphQL](#graphql)
+- [Using Development Version](#using-development-version)
+- [Relative Cursor Pagination](#relative-cursor-pagination)
+- [Limitations](#limitations)
+- [Additional Resources](#additional-resources)
+
+
 ### Getting Started
 #### Public and Custom Apps
 
@@ -119,61 +137,6 @@ _Note: Your application must be public to test the billing process. To test on a
     activated_charge = shopify.ApplicationCharge.find(charge_id)
     has_been_billed = activated_charge.status == 'active'
     ```
-
-### Session tokens
-
-The Shopify Python API library provides helper methods to decode [session tokens](https://shopify.dev/concepts/apps/building-embedded-apps-using-session-tokens). You can use the `decode_from_header` function to extract and decode a session token from an HTTP Authorization header.
-
-#### Basic usage
-
-```python
-from shopify import session_token
-
-decoded_payload = session_token.decode_from_header(
-    authorization_header=your_auth_request_header,
-    api_key=your_api_key,
-    secret=your_api_secret,
-)
-```
-
-#### Create a decorator using `session_token`
-
-Here's a sample decorator that protects your app views/routes by requiring the presence of valid session tokens as part of a request's headers.
-
-```python
-from shopify import session_token
-
-
-def session_token_required(func):
-    def wrapper(*args, **kwargs):
-        request = args[0]  # Or flask.request if you use Flask
-        try:
-            decoded_session_token = session_token.decode_from_header(
-                authorization_header = request.headers.get('Authorization'),
-                api_key = SHOPIFY_API_KEY,
-                secret = SHOPIFY_API_SECRET
-            )
-            with shopify_session(decoded_session_token):
-                return func(*args, **kwargs)
-        except session_token.SessionTokenError as e:
-            # Log the error here
-            return unauthorized_401_response()
-
-    return wrapper
-
-
-def shopify_session(decoded_session_token):
-    shopify_domain = decoded_session_token.get("dest")
-    access_token = get_offline_access_token_by_shop_domain(shopify_domain)
-
-    return shopify.Session.temp(shopify_domain, SHOPIFY_API_VERSION, access_token)
-
-
-@session_token_required  # Requests to /products require session tokens
-def products(request):
-    products = shopify.Product.find()
-    ...
-```
 
 ### Advanced Usage
 It is recommended to have at least a basic grasp on the principles of the [pyactiveresource](https://github.com/Shopify/pyactiveresource) library, which is a port of rails/ActiveResource to Python and upon which this package relies heavily.
--- a/docs/api-access.md
+++ b/docs/api-access.md
@@ -1,0 +1,73 @@
+# Handling access scope operations
+
+#### Table of contents
+
+- [Common ApiAccess operations](#common-apiaccess-operations)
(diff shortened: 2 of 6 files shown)


NabeelAhsen (approved):
Looks good!

.pre-commit-config.yaml
```diff
@@ -6,10 +6,6 @@ repos:
     hooks:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
--   repo: https://github.com/pre-commit/mirrors-autopep8
```
andyw8, line 9: Why did this change?

NabeelAhsen: This comment has more context: https://github.com/Shopify/shopify_python_api/pull/479#discussion_r591842577

paulomarg (approved):
LGTM

shopify/api_access.py
```diff
@@ -0,0 +1,40 @@
...
+    def __str__(self):
+        return self.SCOPE_DELIMITER.join(self._compressed_scopes)
+
+    def __iter__(self):
+        return iter(self._compressed_scopes)
+
+    def __eq__(self, other):
+        return type(self) == type(other) and self._compressed_scopes == other._compressed_scopes
+
+    def __store_scopes(self, scopes):
```
Paulinakhew, line 28: Just out of curiosity what do double leading underscores do? 😅  Is this for name mangling?

Paulinakhew (approved):
Love how detailed the tests are!

shopify/api_access.py
rezaansyed: This is the Python convention of defining "private" methods. Although, methods in Python cannot be truly private. This is just a way to indicate that this method should not be consumed as a public interface as it is more likely to change.

mllemango (approved):

shopify/api_access.py
Paulinakhew: ohh okay makes sense!

shopify/api_access.py
```diff
@@ -0,0 +1,40 @@
...
+        sanitized_scopes = set(list(filter(None, [scope.strip() for scope in scopes])))
```
greatestape, line 29: If we don't want/need the contents of the set to change after we make it, we could make this a `frozenset`.

Also, I don't think you need to convert to `list` before converting to `set`/`frozenset`.

shopify/api_access.py
```diff
@@ -0,0 +1,40 @@
...
+        implied_scopes = set([self.__implied_scope(scope) for scope in sanitized_scopes])
```
greatestape, line 30: again, could be `frozenset` if you don't want this to mutate. Also, I think you can skip the square brackets. It turns it from a list comprehension into a generator, which the set initializer should be happy to consume

shopify/api_access.py
```diff
@@ -0,0 +1,40 @@
...
+        self._compressed_scopes = sanitized_scopes - implied_scopes
+        self._expanded_scopes = sanitized_scopes.union(implied_scopes)
+
+    def __implied_scope(self, scope):
+        match = self.IMPLIED_SCOPE_RE.match(scope)
+        if match:
+            return "{unauthenticated}read_{resource}".format(
+                unauthenticated="" if match.group("unauthenticated") is None else match.group("unauthenticated"),
```
greatestape, line 38: We could use the boolean magic here:

```python
unauthenticated=match.group('unauthenticated') or '',
```

greatestape (commented):
Looks good! I left a bunch of style comments. Nothing profound. I think this PR should include docs in the README.

Paulinakhew:
Pre-commit doesn't actually reformat your code in CI so if you want to reformat stuff locally you can do this
```
pip install -r requirements.txt
pre-commit install
pre-commit run --all-files
```

Paulinakhew:
nicee 🥳

README.md
```diff
@@ -175,6 +175,51 @@ def products(request):
...
+```
+
+##### Checking if ApiAccess covers the access of another
+
+```python
+superset_access = ApiAccess(["write_products", "write_orders", "read_themes"])
+subset_access = ApiAccess(["read_products", "write_orders"])
+
+superset_access.covers(subset_access) # True
+```
```
greatestape, line 221: I think we might want to have something more targeted that illustrates how a developer could leverage this code. Is there code in our POC django app that we could simplify and excerpt here instead?

README.md
```diff
@@ -21,6 +21,24 @@ To easily install or upgrade to the latest release, use [pip](http://www.pip-ins
...
+- [Handling Access Scope Operations](docs/api-access.md)
+- [Advanced Usage](#advanced-usage)
+- [Prefix Options](#prefix-options)
+- [Console](#console)
+- [GraphQL](#graphql)
+- [Using Development Version](#using-development-version)
+- [Relative Cursor Pagination](#relative-cursor-pagination)
+- [Limitations](#limitations)
+- [Additional Resources](#additional-resources)
+
```
NabeelAhsen, line 40: +++++1 for a table of contents!!!


This shows us more clearly that this project's README is close to the "too big for one file" edge

docs/api-access.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Common ApiAccess operations
+
+### Constructing an ApiAccess
+
+```python
+api_access = ApiAccess(["read_products", "write_orders"]) # List of access scopes
+another_api_access = ApiAccess("read_products, write_products") # String of comma-delimited access scopes
+```
+
```
NabeelAhsen, line 20: Could we show a constructor that takes in a string here as well (And show an example of an `unauthenticated` scope)

docs/api-access.md
```diff
@@ -0,0 +1,73 @@
+# Handling access scope operations
+
+#### Table of contents
+
+- [Common ApiAccess operations](#common-apiaccess-operations)
+- [Using ApiAccess to handle changes in app access scopes](#using-apiaccess-to-handle-changes-in-app-access-scopes)
+
+There are common operations that are used for managing access scopes in apps. Such operations include serializing, deserializing and normalizing scopes. Other operations can include checking whether two sets of scopes grant the same API access or whether one set covers the access granted by another set.
```
NabeelAhsen, line 8:
```suggestion
There are common operations that are used for managing [access scopes](https://shopify.dev/docs/admin-api/access-scopes) in apps. Such operations include serializing, deserializing and normalizing scopes. Other operations can include checking whether two sets of scopes grant the same API access or whether one set covers the access granted by another set.
```

Let's link to the resource here for added clarity.

README.md
rezaansyed: Just moved this into a separate doc as discussed.

greatestape (approved):
LGTM!

PR #483, opened 2021-03-12
[Feature] Add sanitize shop param utility

NabeelAhsen:
### Why are these changes introduced?

> Depends on https://github.com/Shopify/shopify_python_api/pull/479

This PR adds a `sanitize_shop_domain` utility method that can be used repeatedly throughout the lifecycle of an app's Oauth with Shopify.

This implementation has been ported from the shopify_app gem: https://github.com/Shopify/shopify_app/blob/17138d7d29d0c85d8f38ace981cac5f726b68833/lib/shopify_app/utils.rb#L4-L14

```py
def sanitize_shop_domain(shop_domain, myshopify_domain="myshopify.com"):
    """Returns a sanitized Shopify shop domain.
    
    If the provided shop domain is invalid or could not be sanitized, returns None.
    
    Parameters:
        shop_domain: str
            A fully-qualified Shopify shop domain or hostname
        myshopify_domain: str, optional
            A custom Shopify domain
        
    Examples:
        >>> print(sanitize_shop_domain('my-shop'))
        'my-shop.myshopify.com'
        
        >>> print(sanitize_shop_domain('bad.com'))
        None
        
        >>> print(sanitize_shop_domain(shop_domain='my-shop', myshopify_domain='myshopify.io ))
        'my-shop.myshopify.io'
    """
````

Developers can use this like so:

```py
from shopify.utils import shop_url
# ...

shop = shop_url.sanitize_shop_domain(params["shop"])
if not shop
    raise ValueError("Bad shop name given")
```

### Things to focus on

1. I'm unsure what type of errors can be thrown by `urllib`
2. Do we encourage Pydocs on our methods? The description written above could have easily been described in the utils file as well.

### Checklist

### DIFF
--- a/shopify/session_token.py
+++ b/shopify/session_token.py
@@ -3,14 +3,13 @@ import re
 import six
 import sys
 
+from shopify.utils import shop_url
+
 if sys.version_info[0] < 3:  # Backwards compatibility for python < v3.0.0
     from urlparse import urljoin
 else:
     from urllib.parse import urljoin
 
-
-HOSTNAME_PATTERN = r"[a-z0-9][a-z0-9-]*[a-z0-9]"
-SHOP_DOMAIN_RE = re.compile(r"^https://{h}\.myshopify\.com/$".format(h=HOSTNAME_PATTERN))
 
 ALGORITHM = "HS256"
 PREFIX = "Bearer "
@@ -69,7 +68,7 @@ def _validate_issuer(decoded_payload):
 def _validate_issuer_hostname(decoded_payload):
     issuer_root = urljoin(decoded_payload["iss"], "/")
 
-    if not SHOP_DOMAIN_RE.match(issuer_root):
+    if not shop_url.sanitize_shop_domain(issuer_root):
         raise InvalidIssuerError("Invalid issuer")
 
 
--- a/shopify/utils/shop_url.py
+++ b/shopify/utils/shop_url.py
@@ -1,0 +1,20 @@
+import re
+import sys
+
+if sys.version_info[0] < 3:  # Backwards compatibility for python < v3.0.0
+    from urlparse import urlparse
+else:
+    from urllib.parse import urlparse
+
+HOSTNAME_PATTERN = r"[a-z0-9][a-z0-9-]*[a-z0-9]"
+
+
+def sanitize_shop_domain(shop_domain, myshopify_domain="myshopify.com"):
+    name = str(shop_domain).lower().strip()
+    if myshopify_domain not in name and "." not in name:
+        name += ".{domain}".format(domain=myshopify_domain)
+    name = re.sub(r"https?://", "", name)
+
+    uri = urlparse("http://{hostname}".format(hostname=name))
+    if re.match(r"{h}\.{d}$".format(h=HOSTNAME_PATTERN, d=re.escape(myshopify_domain)), uri.netloc):
+        return uri.netloc
--- a/test/utils/shop_url_test.py
+++ b/test/utils/shop_url_test.py
@@ -1,0 +1,44 @@
+from shopify.utils import shop_url
+from test.test_helper import TestCase
+
+
+class TestSanitizeShopDomain(TestCase):
+    def test_returns_hostname_for_good_shop_domains(self):
+        good_shop_domains = [
+            "my-shop",
+            "my-shop.myshopify.com",
+            "http://my-shop.myshopify.com",
+            "https://my-shop.myshopify.com",
+        ]
+        sanitized_shops = [shop_url.sanitize_shop_domain(shop_domain) for shop_domain in good_shop_domains]
+
+        self.assertTrue(all(shop == "my-shop.myshopify.com" for shop in sanitized_shops))
+
+    def test_returns_none_for_bad_shop_domains(self):
+        bad_shop_domains = [
+            "myshop.com",
+            "myshopify.com",
+            "shopify.com",
+            "two words",
+            "store.myshopify.com.evil.com",
+            "/foo/bar",
+            "/foo.myshopify.io.evil.ru",
+            "%0a123.myshopify.io ",
+            "foo.bar.myshopify.io",
+        ]
+        sanitized_shops = [shop_url.sanitize_shop_domain(shop_domain) for shop_domain in bad_shop_domains]
+
+        self.assertTrue(all(shop_domain is None for shop_domain in sanitized_shops))
+
+    def test_returns_hostname_for_custom_shop_domains(self):
+        custom_shop_domains = [
+            "my-shop",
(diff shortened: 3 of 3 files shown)


shopify/utils.py
```diff
@@ -72,3 +74,14 @@ def __validate_issuer_and_dest_match(cls, decoded_payload):
 
         if issuer_root != dest_root:
             raise MismatchedHostsError()
+
+
+def sanitize_shop_domain(shop_domain, myshopify_domain="myshopify.com"):
```
rezaansyed, line 79: we shouldn't make `myshopify_domain` configurable as it can allow anyone to override the shopify domain that they are on.

NabeelAhsen: It's configurable in the `shopify_app` gem today - I think it's to accommodate Plus shops (need to verify this)

rezaansyed (approved):
LGTM!

shopify/utils/utilities.py
```diff
@@ -47,8 +49,7 @@ def __validate_issuer(cls, decoded_payload):
 
     @classmethod
     def __validate_issuer_hostname(cls, decoded_payload):
-        hostname_pattern = r"[a-z0-9][a-z0-9-]*[a-z0-9]"
-        shop_domain_re = re.compile(r"^https://{h}\.myshopify\.com/$".format(h=hostname_pattern))
+        shop_domain_re = re.compile(r"^https://{h}\.myshopify\.com/$".format(h=HOSTNAME_PATTERN))
```
paulomarg, line 52: I wonder if it would make sense to have this call `sanitize_shop_domain` and fail if we get an empty result back. That's not necessarily what I would do, I was just wondering whether that made sense.

shopify/utils/utilities.py
```diff
@@ -62,3 +63,14 @@ def __validate_issuer_and_dest_match(cls, decoded_payload):
...
             raise MismatchedHostsError()
+
+
+def sanitize_shop_domain(shop_domain, myshopify_domain="myshopify.com"):
+    name = str(shop_domain).lower().strip()
+    if myshopify_domain not in name and "." not in name:
+        name += ".{domain}".format(domain=myshopify_domain)
+    name = re.sub(r"https?://", "", name)
+
+    uri = urlparse("http://{hostname}".format(hostname=name))
```
paulomarg, line 74: I wonder if we can go with a single RE here:

```suggestion
    matches = re.match(r'(https?://)?({h}\.{d})'.format(h=HOSTNAME_PATTERN, d=myshopify_domain), shop_domain.lower().strip())
    if matches:
        name = matches.group(2)
        return name
```

shopify/utils/utilities.py
NabeelAhsen: That's a good idea!

shopify/utils/utilities.py
paulomarg: I was playing around with this:
```python
print re.match(r'(https?://)?({h})(\.{d})?'.format(h=r"[a-z0-9][a-z0-9-]*[a-z0-9]", d='myshopify.com'), 'https://test.myshopify.com'.lower().strip()).group(1,2,3)
('https://', 'test', 'myshopify.com')
```

```python
print re.match(r'(https?://)?({h})(\.{d})?'.format(h=r"[a-z0-9][a-z0-9-]*[a-z0-9]", d='myshopify.com'), 'https://test'.lower().strip()).group(1,2,3)
('https://', 'test', None)
```

Seems to work! (Note that I've changed the regex slightly from my initial suggestion)

NabeelAhsen: This unfortunately has trouble dealing with all of these bad input cases:

```
"myshop.com",
            "myshopify.com",
            "shopify.com",
            "two words",
            "store.myshopify.com.evil.com",
            "/foo/bar",
            "/foo.myshopify.io.evil.ru",
            "%0a123.myshopify.io ",
            "foo.bar.myshopify.io",
```

NabeelAhsen: Hmm yeah the regexes above fail for the following inputs: 
* `shopify.com`: returns `shopifymyshopify.com` (expected `None`)
* `myshop.com`: returns `myshopmyshopify.com` (expected `None`)
* `two words`: returns `twomyshopify.com` (expected `None`)
* `foo.bar.myshopify.io`: returns `foomyshopify.com` (expected `None`)

paulomarg: Ah yes, my regex didn't account for cases where the `.` in the domain isn't the start for `.myshopify.com`. That requirement is contextual, so we probably won't be able to handle it with a straight-up regexp. We could rewrite this code around a regex, but I think it would end up looking pretty much the same. LGTM!

paulomarg (approved):

test/utils/utilities_test.py
```diff
@@ -85,3 +85,43 @@ def test_returns_decoded_payload(self):
...
+    def test_returns_hostname_for_good_shop_domains(self):
+        good_shop_domains = [
+            "my-shop",
+            "my-shop.myshopify.com",
+            "http://my-shop.myshopify.com",
+            "https://my-shop.myshopify.com",
+        ]
+        sanitized_shops = [sanitize_shop_domain(shop_domain) for shop_domain in good_shop_domains]
+
+        self.assertTrue(all(shop_domain is not None for shop_domain in sanitized_shops))
```
Paulinakhew, line 100: should we replace this line with what we expect to be returned?

test/utils/utilities_test.py
```diff
@@ -85,3 +85,43 @@ def test_returns_decoded_payload(self):
...
+    def test_returns_hostname_for_custom_shop_domains(self):
+        custom_shop_domains = [
+            "my-shop",
+            "my-shop.myshopify.io",
+            "http://my-shop.myshopify.io",
+            "https://my-shop.myshopify.io",
+        ]
+        sanitized_shops = [sanitize_shop_domain(shop_domain, "myshopify.io") for shop_domain in custom_shop_domains]
+
+        self.assertTrue(all(shop_domain is not None for shop_domain in sanitized_shops))
```
Paulinakhew, line 127: same with this

test/utils/utilities_test.py
NabeelAhsen: Does the latest change help clarify the expected output?

`self.assertTrue(all(shop == "my-shop.myshopify.com" for shop in sanitized_shops))`
and 
`self.assertTrue(all(shop == "my-shop.myshopify.io" for shop in sanitized_shops))`

Paulinakhew: yeah that's perfect!

Paulinakhew (approved):
cool! 👌

mllemango (approved):
v cool

PR #489, opened 2021-03-17
[Feature]: Add access scopes in Session object

rezaansyed:

We need to store access scopes in the `Session` object so that apps have access to the scope granted to them after OAuth. These scopes can then be used by apps to keep track of scopes granted.

Storing `access_scopes` as an instance variable in the `Session` object on construction and after access token exchanges.

### Checklist

### DIFF
--- a/shopify/session.py
+++ b/shopify/session.py
@@ -10,6 +10,7 @@ except ImportError:
 import re
 from contextlib import contextmanager
 from six.moves import urllib
+from shopify.api_access import ApiAccess
 from shopify.api_version import ApiVersion, Release, Unstable
 import six
 
@@ -45,10 +46,11 @@ class Session(object):
         yield
         shopify.ShopifyResource.activate_session(original_session)
 
-    def __init__(self, shop_url, version=None, token=None):
+    def __init__(self, shop_url, version=None, token=None, access_scopes=None):
         self.url = self.__prepare_url(shop_url)
         self.token = token
         self.version = ApiVersion.coerce_to_version(version)
+        self.access_scopes = access_scopes
         return
 
     def create_permission_url(self, scope, redirect_uri, state=None):
@@ -72,7 +74,10 @@ class Session(object):
         response = urllib.request.urlopen(request)
 
         if response.code == 200:
-            self.token = json.loads(response.read().decode("utf-8"))["access_token"]
+            json_payload = json.loads(response.read().decode("utf-8"))
+            self.token = json_payload["access_token"]
+            self.access_scopes = json_payload["scope"]
+
             return self.token
         else:
             raise Exception(response.msg)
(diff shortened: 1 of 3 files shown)


shopify/session.py
```diff
@@ -10,6 +10,7 @@
 import re
 from contextlib import contextmanager
 from six.moves import urllib
+from shopify.api_access import ApiAccess
```
NabeelAhsen, line 13: I think all that's needed here is the following:

```suggestion
from shopify import ApiAccess
```

The base shopify `__init__.py` file specifies `from api_access import *`.

rezaansyed: It fails locally as it fails to find the module.

test/session_test.py
```diff
@@ -257,6 +258,32 @@ def test_raise_error_if_timestamp_is_too_old(self):
...
+            shop_url="testshop.myshopify.com",
+            version="unstable",
+            token="any-token",
+            access_scopes="read_products, write_orders",
+        )
+
+        expected_access_scopes = shopify.ApiAccess("read_products, write_orders")
+        self.assertEqual(expected_access_scopes, session.access_scopes)
+
+    def test_access_scopes_when_scopes_of_apiaccess_type_passed_in(self):
```
NabeelAhsen, line 276: nit:

I had to re-read this test name (and then the code) to understand what it meant, could we simplify the name of this test?

NabeelAhsen (approved):
Minor non-blocking comments. LGTM

shopify/session.py
NabeelAhsen: Does the pytest runner fail?

NabeelAhsen: Oh it might be `from .api_access import ApiAccess`. What does the failure say specifically?

rezaansyed:
```
from shopify import ApiAccess
E   ImportError: cannot import name ApiAccess
```

rezaansyed: `from .api_access import ApiAccess` works but idk if it might be better to be more explicit and keep it as `from shopify.api_access import ApiAccess`

greatestape (approved):
LGTM

test/session_test.py
```diff
@@ -257,6 +258,32 @@ def test_raise_error_if_timestamp_is_too_old(self):
             session = shopify.Session("http://localhost.myshopify.com", "unstable")
             session = session.request_token(params)
 
+    def test_access_scopes_are_nil_by_default(self):
+        session = shopify.Session("testshop.myshopify.com", "unstable", "any-token")
+        self.assertIsNone(session.access_scopes)
+
+    def test_access_scopes_when_valid_scopes_passed_in(self):
```
paulomarg, line 265: Out of curiosity, what will happen if invalid scopes are passed in? Is it worth testing that specifically or can we just trust the ApiAccess tests?

paulomarg (approved):
Makes sense to me!

test/session_test.py
NabeelAhsen: There is some test coverage in this file: https://github.com/Shopify/shopify_python_api/blob/455bd5a94d2d29d005798f1b81ba4e3fd816672e/test/api_access_test.py#L13-L17

From the library's perspective, the `ApiAccess` object will raise this error. If a bad access scope is passed in to the permission_url during OAuth, Shopify will elegantly let you know during the redirect dance.

rezaansyed: I think we can trust the `ApiAccess` tests in here. `ApiAccess` will throw an `ApiAccessError` if that fails. That is already covered in `ApiAccess` tests.

PR #499, opened 2021-03-26
[Bug fix] - Return None when sanitizing shop domain if None passed in

NabeelAhsen:

> **Bug:** `shop_url.sanitize_shop_domain(None)` returns `"none.myshopify.com"`
> * I just found out that `str(None) == "none"` in Python 😬 

~~This PR is intended to be a patch for shopify_python_api `v8.3`~~. It introduces a check to see if `shop_domain` is `None`.

### Releasing

~~I noticed that we've released `v8.4` already. Ideally, I'd like to release this as a patch for both `v8.4` and `v8.3.2`. Do we have a guide on how to do this?~~

This is intended to be a patch release.

### Checklist

### DIFF
--- a/shopify/utils/shop_url.py
+++ b/shopify/utils/shop_url.py
@@ -10,7 +10,7 @@ HOSTNAME_PATTERN = r"[a-z0-9][a-z0-9-]*[a-z0-9]"
 
 
 def sanitize_shop_domain(shop_domain, myshopify_domain="myshopify.com"):
-    name = str(shop_domain).lower().strip()
+    name = str(shop_domain or "").lower().strip()
     if myshopify_domain not in name and "." not in name:
         name += ".{domain}".format(domain=myshopify_domain)
     name = re.sub(r"https?://", "", name)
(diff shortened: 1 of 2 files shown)


mllemango (approved):
Hahaha TIL
Neat use of `or`

NabeelAhsen:
@mllemango  Do we have a release strategy to patch this for release `v8.3`?

andyw8:
We generally don't patch older release, what would prevent those on v8.3 from updating?

NabeelAhsen:
> We generally don't patch older releases, what would prevent those on v8.3 from updating?

Oh good point, nothing's breaking I suppose

:shipit:

NabeelAhsen:
I'm updating this PR to include a Changelog update + patch version bump as well

mllemango:
Should we update the title of the PR to be Release v8.4.1 in this case?

andyw8:
We should be using separate PRs for the changes and the releases.

NabeelAhsen:
I've updated this PR to only include the bug fix, I'll draft a follow-up patch release in a separate PR

rezaansyed (approved):
LGTM!

PR #504, opened 2021-04-01
Fix API update reminder action

andyw8:
This was failing with:

https://github.com/Shopify/shopify_python_api/actions/runs/707075633

`end of the stream or a document separator is expected at line 2, column 12: assignees: @platform/dev-tools-education ^`


andyw8:
(also the team name was incorrect!)

PR #507, opened 2021-04-01
Remove assignees field

andyw8:
Assigning to a team seems to be problematic, so for now we'll manually assign these when created.


andyw8:
You can't actually can't assign a team to an issue: https://github.com/dear-github/dear-github/issues/170

paulomarg (approved):

PR #510, opened 2021-04-13
Add a sample Django app to README.md

NabeelAhsen:

Now that we're officially support session tokens in this library, we'd like to share a sample app that we've developed using this library.

* Adds a `Sample apps built using this library` section in the README
* Links to a Shopify-maintained sample app that was built using Django and shopify_python_api

### Checklist


andyw8:
Not a blocker, but would it be possible to add a basic test for that sample app, and enable Dependabot for `shopify_python_api`, so that we can ensure it remains valid?

PR #511, opened 2021-04-14
[WIP] [Feature] Add some basic logging around request/response process. 

### LINKED ISSUE #500: Validate Kwargs before and possibly after passing them to Shopify
## Overview

Validate kwargs before passing them to shopify and either raise an exception or warn if the kwarg is likely to produce no effect (the current behavior). I'd propose implementing this as a session-level flag to change behavior which would likely be implemented in the   __encoded_params_for_signature method of the Session class. 

something like:
```
with shopify.Session.temp(shop_url, api_version, token, validate_kwargs='fail'):
   orders = shopify.Order.find(ids="1234")
```
where:
default - current behavior
warn - current behavior, with warning that kwargs will not work as expected
fail - raise exception before generating query string. 

With this new behavior:

```
shopify.Order.find(ids="1234")
##################
InvalidKwargError: The ids kwarg should be a comma-separated list of ids.
```

...

## Type

- [x] New feature
- [x] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing?

I recently had the experience of trying to fetch two orders while testing some code: 

`shopify.Order.find(ids="1234,5678")`
Next I wanted to check for a single order so I deleted the second order:
`shopify.Order.find(ids="1234")`

Those with a keen eye could spot the bug: ID's needs a _comma-separated list of order IDs._ according to the [docs](https://shopify.dev/docs/admin-api/rest/reference/orders/order#index-2021-01). Without the comma, the kwarg is basically disregarded and all (non-archived) orders are returned. This can lead to unexpectedly long running queries that is hard to debug when this code is generated dynamically or buried under several layers of other code.

Here are a few examples of the proposed validation: 

 - IDS : if "ids" in kwargs, validate that there is at least one "," and that the values in the "ids" string are integers. The response could also be validated as we'd expect to receive <= number of objects in the response. eg: if the "ids" filter has two elements, we can expect <=2 objects in the response. 
 - status: check that the status kwarg has one of the valid values ['open', 'closed', 'cancelled', 'any']
 - created_at_min: check that the min-date is a valid iso-formatted date 
 
With the warn/fail arguments I proposed above, if the kwarg didn't pass the test, the code would loudly warn or fail and make debugging much easier. 

An alternative, and possibly easier to implement solution, would be to add configurable logging. The ability to easily see the full query string as well as the raw response body could make the same debugging a lot easier. 

I'm happy to open a PR for either proposed solution, but would appreciate some guidance from the maintainers before doing so. 
...

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

---

## Checklist

- [x] I have described this feature request in a way that is actionable (if possible)

samLozier:
Currently, problematic queries are difficult to debug, this should make slightly it easier.

PR is opened to address this discussion: https://github.com/Shopify/shopify_python_api/issues/500

Fixes #500 

  Context about the problem that’s being addressed: 

 Adds some very rudimentary debug level logging in the session and base files. 

Further discussion: 
- I used f-strings, however I noticed that tox is still running tests for 2.7 and <3.7 all of which are EOL or will be shortly. Can we drop that support and move to currently supported python versions? If old python needs to be supported, I can switch them to something pre f-string. 
- this is flagged as a WIP because I know people will probably have strong feeling around the verbosity and format of the logs as well as the method for toggling them. We could use a -v flag/verbosity kwarg, the standard logging conventions, or really anything else that makes sense. 

### Checklist

### DIFF
--- a/shopify/session.py
+++ b/shopify/session.py
@@ -13,6 +13,9 @@ from six.moves import urllib
 from shopify.api_access import ApiAccess
 from shopify.api_version import ApiVersion, Release, Unstable
 import six
+import logging
+
+log = logging.getLogger(__name__)
 
 
 class ValidationException(Exception):
@@ -71,10 +74,13 @@ class Session(object):
         url = "https://%s/admin/oauth/access_token?" % self.url
         query_params = dict(client_id=self.api_key, client_secret=self.secret, code=code)
         request = urllib.request.Request(url, urllib.parse.urlencode(query_params).encode("utf-8"))
+        log.debug(f"Request URL: {url}, query params: {query_params}")
         response = urllib.request.urlopen(request)
+        logging.debug(f"Response: {response}")
 
         if response.code == 200:
             json_payload = json.loads(response.read().decode("utf-8"))
+            log.debug(f"Response: {json_payload}")
             self.token = json_payload["access_token"]
             self.access_scopes = json_payload["scope"]
 
--- a/shopify/base.py
+++ b/shopify/base.py
@@ -7,9 +7,10 @@ import threading
 import sys
 from six.moves import urllib
 import six
-
+import logging
 from shopify.collection import PaginatedCollection
 from pyactiveresource.collection import Collection
+log = logging.getLogger(__name__)
 
 # Store the response from the last request in the connection object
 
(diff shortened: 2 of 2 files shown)


samLozier:
@mllemango Thanks, I didn't realize that and will take a look. I suppose the question might be moot if I end up using the pyactiveresource logging, but any thoughts on the <3.7 compatibility question?

mllemango:
Oh yes, sorry, fully support dropping pre 3.7 compatibility

shopify/session.py
```diff
@@ -71,10 +74,13 @@ def request_token(self, params):
         url = "https://%s/admin/oauth/access_token?" % self.url
         query_params = dict(client_id=self.api_key, client_secret=self.secret, code=code)
         request = urllib.request.Request(url, urllib.parse.urlencode(query_params).encode("utf-8"))
+        log.debug(f"Request URL: {url}, query params: {query_params}")
         response = urllib.request.urlopen(request)
+        logging.debug(f"Response: {response}")
```
slyapustin, line 79: I think you want to use `log.debug(...)` here instead.

PR #518, opened 2021-05-03
Add private python index

mllemango:

Adding Shopify's python index

### Checklist

- [ ] ~I have updated the CHANGELOG (if applicable)~

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,9 @@
 version: 2
+registries:
+  python-shopify:
+    type: python-index
+    url: https://pkgs.shopify.io/basic/data/python/simple
+    token: ${{secrets.PYTHON_CLOUDSMITH_TOKEN}}
 updates:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -1,4 +1,9 @@
 version: 2
+registries:
+  python-shopify:
+    type: python-index
+    url: https://pkgs.shopify.io/basic/data/python/simple
+    token: ${{secrets.PYTHON_CLOUDSMITH_TOKEN}}
```
pfeffer, line 6: this looks good! you  also need to add `registries` key under `updates` like here https://github.com/Shopify/shopify/blob/master/.github/dependabot.yml#L27 so it's actually being used for that package-ecosystem

pfeffer: oh, so sorry! it was the last line and i totally missed it 🙈

PR #540, opened 2021-10-05
Fixes doc link to Session Tokens

dlindquist:
The link needs to include a ".md" suffix. 

https://github.com/Shopify/shopify_python_api/blob/master/docs/session-tokens -> https://github.com/Shopify/shopify_python_api/blob/master/docs/session-tokens.md


andyw8:
Thanks @dlindquist!

PR #543, opened 2021-10-07
Suggestion for supporting ongoing API version releases

sinejoe:
Be less rigid with API versions. Having to manually update this repo every three months is getting cumbersome it seems and we are lagging behind by weeks in getting the new versions supported as they are released by Shopify. This repo doesn't need to strictly enforce the deprecations and new releases. Developers can still make a request to the 2019-04 api version successfully so there isn't a major need IMO to block deprecated API versions if a developer chooses to use. Further, each new release and the next release candidate should be usable/testable here without having to override.

Addresses original issue #[339](https://github.com/Shopify/shopify_python_api/issues/339) for which a fix was attempted in #[441](https://github.com/Shopify/shopify_python_api/pull/441) but caused more problems than solved. 

Replacing a fixed, hard-coded list of supported/valid API versions with an allowable range of all API versions plus new ones automatically going forward including next release candidate.

### Checklist

I did not update these items awaiting acceptance or review/discussion of this PR.

### DIFF
--- a/shopify/api_version.py
+++ b/shopify/api_version.py
@@ -32,8 +32,8 @@ class ApiVersion(object):
     def define_known_versions(cls):
         cls.define_version(Unstable())
         # Generate quarterly API versions thru next release candidate
-        ver_yr = int(MIN_API_VERSION.split('-')[0])
-        ver_mo = int(MIN_API_VERSION.split('-')[1])
+        ver_yr = int(MIN_API_VERSION.split("-")[0])
+        ver_mo = int(MIN_API_VERSION.split("-")[1])
         now = datetime.utcnow()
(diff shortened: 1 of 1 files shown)


mllemango (commented):
I like this a lot 
Would you mind adding unit tests?

mllemango:
Hey @sinejoe, we're currently looking into alternative ways of automating these releases that would apply to all our libraries and it seems like we're leaning towards a more rigid control of api versions rather than what's being suggested here. 

Thanks for putting forth this change, it sparked a lot of conversation about how we want to proceed with these updates!

With all that said, I'm going to close this PR, feel free to open it again if anything else comes up.

PR #562, opened 2022-01-05
Reintroduce older api version to prevent breaking change upon release

mllemango:

Taking out the older version would make our next release a breaking change and we dont want a new major version released 4 times a year.

### Checklist


thecodepixi:
I wonder if continuing to store the old API versions in the library will lead to a false impression that those versions are supported? Because attempting to use an unsupported version could definitely still break your app.

mllemango:
Closing, letting stale versions stay is working against best practices and these changes are ultimately breaking.

PR #571, opened 2022-02-25
Add operation_name parameter to Graphql.execute

JJSphar:
When constructing non-trivial GraphQL queries, it can be helpful to have multiple NamedOperations in the same document. 
For example, I could have a fragment which I'd like to reuse between two queries:

```graphql
    fragment orderInfo on Order {
        id
        name
        createdAt
    }
    
    query GetOneOrder($order_id: ID!){
        node(id: $order_id){
            ...OrderInfo
        }
    }
    
    query GetManyOrders($order_ids: [ID]!){
        nodes(ids: $order_ids){
           ...OrderInfo
        }
    }
```

From [the GraphQL docs](https://graphql.org/learn/serving-over-http/#post-request), an "operationName" parameter can be included in the POST request to specify which operation should be performed. 

When sending a manual GraphQL POST request to Shopify, the "operationName" parameter is accepted / respected. 

**Currently**,  
There is no way to specify this operationName as a POST parameter when using the `execute()` method from Shopify's `GraphQL` helper client. 

The body of the post request is being constructed [here](https://github.com/Shopify/shopify_python_api/blob/master/shopify/resources/graphql.py#L22).

This PR adds an `operation_name` parameter to the `execute()` method in question. The value will added to the body of POST request, under the "operationName" key. 

The presence of `operationName` in the request is handled by the Shopify GraphQL server, which will execute the correct operation and return the results as normal.

The README has also been updated with this example of a more complex GraphQL query, using both the new `operation_name` and the existing `variables` parameters.

### Checklist

- [X] I have updated the CHANGELOG (if applicable)
  -  This doesn't seem big enough to update the CHANGELOG, but I'd be happy to do so if a maintainer disagrees.
- [X] I have followed the [Shopify Python](https://github.com/Shopify/shopify_python) guide

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,9 +183,9 @@ This library also supports Shopify's new [GraphQL API](https://help.shopify.com/
 result = shopify.GraphQL().execute('{ shop { name id } }')
 ```
 
-You can perform more complex operations using the `variables` and `operation_name` parameters of `execute`. 
+You can perform more complex operations using the `variables` and `operation_name` parameters of `execute`.
 
-For example, this GraphQL document uses a fragment to construct two named queries - one for a single order, and one for multiple orders:  
+For example, this GraphQL document uses a fragment to construct two named queries - one for a single order, and one for multiple orders:
 
 ```graphql
     # ./order_queries.graphql
@@ -195,13 +195,13 @@ For example, this GraphQL document uses a fragment to construct two named querie
         name
         createdAt
     }
-    
+
     query GetOneOrder($order_id: ID!){
         node(id: $order_id){
             ...OrderInfo
         }
     }
-    
+
     query GetManyOrders($order_ids: [ID]!){
         nodes(ids: $order_ids){
            ...OrderInfo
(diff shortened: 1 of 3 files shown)


JJSphar:
I've just signed the CLA.

Let me know if there are any questions or concerns with this addition, or my approach.

Thanks!

README.md
```diff
@@ -183,6 +183,45 @@ This library also supports Shopify's new [GraphQL API](https://help.shopify.com/
...
 ```
 
+You can perform more complex operations using the `variables` and `operation_name` parameters of `execute`.
+
+For example, this GraphQL document uses a fragment to construct two named queries - one for a single order, and one for multiple orders:
+
+```graphql
+    # ./order_queries.graphql
+
+    fragment orderInfo on Order {
```
mllemango, line 193:
```suggestion
    fragment OrderInfo on Order {
```

JJSphar: We should be good to go now! For some reason it didn't show up until I did a pull + push.

PR #614, opened 2022-10-28
Update API version to 2022-10

Alchez:
Fixes #611 

### Checklist

- [x] I have updated the CHANGELOG (if applicable)
- [x] I have followed the [Shopify Python](https://github.com/Shopify/shopify_python) guide

PR #624, opened 2022-12-12
Fix triggering CI for open source PRs

dylanahsmith:
## Problem

I noticed in https://github.com/Shopify/shopify_python_api/pull/614 that the CI github action wasn't run.  That github action workflow has `on: [push]` for the event that triggers it, but that would only happen for pushes directly onto the Shopify/shopify_python_api repo, so wouldn't happen for a fork (e.g. Alchez/shopify_python_api in that issue).

## Solution

Add the pull_request event, so it will be triggered when a PR is opened.

This workflow doesn't require access to secrets, which [github encrypted secrets document says](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow) "are not passed to the runner when a workflow is triggered from a forked repository."

### DIFF
--- a/shopify/base.py
+++ b/shopify/base.py
@@ -17,9 +17,6 @@ from pyactiveresource.collection import Collection
 class ShopifyConnection(pyactiveresource.connection.Connection):
     response = None
 
-    def __init__(self, site, user=None, password=None, timeout=None, format=formats.JSONFormat):
-        super(ShopifyConnection, self).__init__(site, user, password, timeout, format)
-
     def _open(self, *args, **kwargs):
         self.response = None
         try:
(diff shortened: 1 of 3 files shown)


paulomarg (approved):
I think running this on `pull_request` makes sense, but it seems that CI is still failing 🤔.

shopify/base.py
```diff
@@ -17,9 +17,6 @@
 class ShopifyConnection(pyactiveresource.connection.Connection):
     response = None
 
-    def __init__(self, site, user=None, password=None, timeout=None, format=formats.JSONFormat):
```
paulomarg, line 20: These got flagged as linting errors because they replicate the parent's signature, so they're not changing anything.

Issue #634, opened 2023-02-04
Incorrect URL generated when creating fulfillment records - API version 2023-01


iCodeTechnologies:
# Issue summary

When calling shopify.Fulfillment.create the resulting URL is incorrect preventing the successful creation of the Fulfillment record.

## Expected behavior

The URL generated by the Shopify client library should be 

```/admin/api/<API_VERSION>/fulfillments.json```

## Actual behavior

The URL generated by the Shopify client library is

```/admin/api/<API_VERSION>/orders//fulfillments.json```

_**NOTE:** The call to shopify.Fulfillment.create does not accept an order_id parameter._

## Steps to reproduce the problem

1. Create Shopify session with api_version set to '2023-01'
2. Call shopify.Fulfillment.create with a valid fulfillment payload

## Reduced test case

```
import logging
import shopify

logging.basicConfig(level=logging.DEBUG)

shop_uri = 'xxx.myshopify.com'
api_version = '2023-01'
access_token = 'shppa_xxx'

session = shopify.Session(shop_uri, api_version, access_token)
shopify.ShopifyResource.activate_session(session)

payload = dict(
    line_items_by_fulfillment_order=[dict(fulfillment_order_id="5894084460693", fulfillment_order_line_items=[
        dict(id="12190922277013", quantity="1")])])

shopify.Fulfillment.create(payload)

```

The best way to get your bug fixed is to provide a [reduced test case](https://developer.mozilla.org/en-US/docs/Mozilla/QA/Reducing_testcases).

---

## Checklist

- [x] I have described this issue in a way that is actionable (if possible)

**Comment by iCodeTechnologies** [comment]
Looking in the code I can see that the Order resource implements a function called _prefix that appears to allow for the URL to contain a path element "/customers" if a customer_id is included in the options, or omit the "/customers" path element in no customer_id is present.

```
class Order(ShopifyResource, mixins.Metafields, mixins.Events):
    _prefix_source = "/customers/$customer_id/"

    @classmethod
    def _prefix(cls, options={}):
        customer_id = options.get("customer_id")
        if customer_id:
            return "%s/customers/%s" % (cls.site, customer_id)
        else:
            return cls.site
```

would this approach work for the Fulfillment resource?

bjquinn:
> ```
> shopify.Fulfillment._prefix_source = ''
> ```

[...]

PR #635, opened 2023-02-06
[WIP] [FIX] Fix URL generation for Fulfillment

iCodeTechnologies:

Fix Fulfillment.py to generate the correct URL for API 2023-01.

If no order_id supplied, URL should be /admin/api/<api_version>/fulfillments.py

currently the library produces the URL

/admin/api/<api_version>/orders//fulfillments.py

See issue #634

Modifying the code in Fulfillment that generates the URL so that it complies with the requirements for API version 2023-01

**NOTE:** The added code is based on code already in the library for other resource classes
### Checklist

### DIFF
--- a/shopify/resources/fulfillment.py
+++ b/shopify/resources/fulfillment.py
@@ -5,6 +5,14 @@ import json
 class Fulfillment(ShopifyResource):
     _prefix_source = "/orders/$order_id/"
 
+    @classmethod
+    def _prefix(cls, options={}):
+        order_id = options.get("order_id")
+        if order_id:
+            return "%s/orders/%s" % (cls.site, order_id)
+        else:
+            return cls.site
+    
     def cancel(self):
         self._load_attributes_from_response(self.post("cancel"))
 


MohamedAliHamza:
Hi @iCodeTechnologies, did you test it? because seems to me it consumes the same endpoint, but we need to consume this [one](https://shopify.dev/docs/api/admin-rest/2023-01/resources/fulfillment#post-fulfillments). which will require changes in the request body I think.

remibergsma:
@MohamedAliHamza This PR solves an issue for me where the url generated by the library is incorrect (when no order_id is given): `/admin/api/<api_version>/orders//fulfillments.py`
(Double slash) This results in a 403. With this PR, the URL is correct:
`/admin/api/<api_version>/orders/fulfillments.py`
And the request is successful.

I run into this when migrating my code to the new fulfillment options, where the old ones are not working after 2023-04-01. Did some searing and found that this PR solves the issue I had.

iCodeTechnologies:
> Hi @iCodeTechnologies, did you test it? because seems to me it consumes the same endpoint, but we need to consume this [one](https://shopify.dev/docs/api/admin-rest/2023-01/resources/fulfillment#post-fulfillments). which will require changes in the request body I think.

Hi Mohamed, I tested the pull request with my application; it solves the issue I was having, which is that the Fulfillment POST URL was not correct 
`/admin/api/<api_version>/orders//fulfillments.py`
instead of
`/admin/api/<api_version>/fulfillments.py`

Of course, other changes are required when moving to the latest API.

This is a simplified extract of what my application is doing to send fulfillments back to Shopify; in the real system I compare the line_items against the orders in my system etc

It should give you an idea of how you can do fulfillment with FulfillmentOrders.

`
fulfillment_orders = []
fulfillment_orders = get_all_resources(shopify.FulfillmentOrders, limit=250, order_id=XXXXXX)

payload = {}
for fulfillment_order in fulfillment_orders:
    fulfillment_order_item = dict(fulfillment_order_id=str(fulfillment_order.attributes['id']),
                                          fulfillment_order_line_items=[])

    for fulfillment_line_item in fulfillment_order.attributes['line_items']:
        fulfillment_order_item['fulfillment_order_line_items'].append(
                        dict(id=str(fulfillment_line_item.attributes['id']),
                             quantity=str(fulfillment_line_item.attributes['quantity'])))

payload['notify_customer'] = True
payload['line_items_by_fulfillment_order'] = []
payload['line_items_by_fulfillment_order'].append(fulfillment_order_item)
shopify.Fulfillment.create(payload)
`

Cheers
Dan

markm-catalyte:
@MohamedAliHamza , have you tried using FulfillmentV2 in the same file?

pama1999:
why is the python library from shopify so dead?

Issue #725, opened 2024-07-08
Add support for Python 3.12


PacificGilly:
# Issue summary

In Python 3.12, there are numerous changes low-level changes that were implemented. The most important was [PEP-632](https://peps.python.org/pep-0632/). This seems to impact transient packages like [PyYaml](https://github.com/yaml/pyyaml/issues/756) for versions <=6.0.0 (see below for stacktrace). By forcing the PyYaml dependency for PY3.12 implementations to use 6.0.1 or later resolves the issue.

## Expected behavior

Installing the shopify API library builds successfully

## Actual behavior

The build fails to install if its coupled with PyYaml 6.0.0 or earlier. The following stacktrace is raised:
```bash
  ChefBuildError

  Backend subprocess exited when trying to invoke get_requires_for_build_wheel
  
  running egg_info
  writing lib/PyYAML.egg-info/PKG-INFO
  writing dependency_links to lib/PyYAML.egg-info/dependency_links.txt
  writing top-level names to lib/PyYAML.egg-info/top_level.txt
  Traceback (most recent call last):
    File "/home/circleci/.local/share/pypoetry/venv/lib/python****/site-packages/pyproject_hooks/_in_process/_in_process.py", line 353, in <module>
      main()
    File "/home/circleci/.local/share/pypoetry/venv/lib/python****/site-packages/pyproject_hooks/_in_process/_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/circleci/.local/share/pypoetry/venv/lib/python****/site-packages/pyproject_hooks/_in_process/_in_process.py", line 118, in get_requires_for_build_wheel
      return hook(config_settings)
             ^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/tmplgj9ejjh/.venv/lib/python****/site-packages/setuptools/build_meta.py", line 325, in get_requires_for_build_wheel
      return self._get_build_requires(config_settings, requirements=['wheel'])
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/tmp/tmplgj9ejjh/.venv/lib/python****/site-packages/setuptools/build_meta.py", line 295, in _get_build_requires
      self.run_setup()
[...]

PR #726, opened 2024-07-09
Fix(725): Add support for building the package in Python 3.12

PacificGilly:
### Description
As discussed in the [issue](https://github.com/Shopify/shopify_python_api/issues/725), there is an issue when installing this package using Python 3.12. The problem is caused by
[PEP-632](https://peps.python.org/pep-0632/) removes `distutils` from the built-in libraries this version of Python is shipped with. In versions 6.0.0 and lower of PyYaml, this causes the build to fail. As mentioned in their [issue](https://github.com/yaml/pyyaml/issues/756), using v6.0.1 or later will resolve this issue. So, we are updating our `setup.py` install_requires dependency specifications to reflect this.

While I'm here, I've also updated the classifiers to include support for PY3.12 and show this more clearly in the repo's README.

### Checklist


PacificGilly:
I have signed the CLA!

PR #753, opened 2025-01-09
drop "six" usage

a-detiste:
Hi,

shopifiy has enterred Debian today https://tracker.debian.org/pkg/python-shopifyapi :-)

... but we are slowly removing old Python2 compatibility layers, such as `six`
https://wiki.debian.org/Python3-six-removal

Please consider this PR.

The project metadata already state that Py2.7 support is dropped; that's why I spent the time to preprase this PR.


sle-c:
hi @a-detiste , thank you for your valuable contribution. We'll review your PR as soon as we can.

